PR #415, opened 2015-01-06
Adapted german language, unified a lot of phrases

st-anton:
Run through the language files:
- unified duplicate entries that gives the same output
- replaced phrases, that are alredy present in the Joomla standard language file (en-GB.ini)
- deleted keys, that are present in the Joomla standard language
- removed phrases, that are no more used (template with tables ...)


tomfuller2:
Thank you so much for doing this work!

bcordis:
Merge per label.

PR #424, opened 2015-01-10
[work-in-progress] Feature: added composer instead of pear

mheiniger:
As the future goes away from pear and in direction of composer (and is wasnt able to install the correct versions with pear on my ubuntu machine), i started to use composer for for development.

The big plus: its less depending on whats installed on a machine, because it installs everything inside the "vendor" folder of the project.

Maybe it can be a first step in that direction.
On development branch i tried it too, but at the current state i can't get it running, maybe it needs different packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ We appreciate contributions in varioius capacities, below are some ways that you
 2. [Create a topic branch.][branch]
 3. Implement your feature or bug fix.
 4. If you implemented a new feature or added an extra functionality, create/update unit tests for that feature
-4. Run `phing build` (requires [Phing][phing] to be installed)
+4. Run `composer install --dev && vendor/bin/phing build` (requires [Composer][https://getcomposer.org] to be installed)
 5. If not building sucessfully, go back to step **3**
 6. Add your files to repositiory: `git add .`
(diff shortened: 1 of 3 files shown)


bcordis:
Until you can get a stable version of composer working I'll keep the pull open.

We right now use phing to build the packages and move test data around.

bcordis:
When you are ready for us to review the pull request pleas add Labile 1-ready

bcordis:
@mheiniger do you still want to do this pull or are you good with phing now? we have been using it for a quite a will.

Let me know what you are thinking of?

mheiniger:
Its not about removing phing, its about removing the need of globally installing pear-packages. Phing will then be installed with composer into the vendors directory.

bcordis:
Just let me know when you have a tested system. Then I'll test before putting it into dev.

mheiniger:
Yes i will.. the problem is, that the phpunit tests on your development branch fail, so before that is fixed i never know if the errors are because of my changes or because of the problems of the tests of the development branch.

bcordis:
We never worked up phpunit yet so if you know it well you can start working it up. Have not had time. Let me know if you need help there.

mheiniger:
This pull-request is on the wrong branch anyways. I will close it and open one for the development branch.

PR #434, opened 2015-01-18
Test-dependencies with composer

mheiniger:
Update Phing to local installed packages with "Composer" instead of globally installed pear packages.
Makes the building process more operating-system independent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,5 @@ tests/unit/.joomla-cms/
 /bin/phploc
 /bin/phpmd
 /bin/phpunit
+/bin/phpcs
 /composer.phar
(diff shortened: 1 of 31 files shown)


bcordis:
Not attached to any milestone as this is independent of the component working.

mheiniger:
Yes.. is something like "infrastructure work" for developers to have an easier ramp-up

bcordis:
Put ready label when you have a test version.

mheiniger:
I think its now ready for you to review. All phing tasks of `phing build` work now, including phpunit.

The added files in the folder tests/unit/core/ are taken from the Joomla 3.3.6, and also the phpcs ruleset is completed from there.

bcordis:
@genu Can you also take a look at this. From what i can see it look really good. I would like to merge this by are next meeting.

Issue #450, opened 2015-02-05
Frontend: Administration pages need Joomla admin language file


st-anton:
In the Frontend administration pages for studies and media files (and I guess for comments as well), a lot of Joomla language phrases like JTOOLBAR_PUBLISH are used. Please add
`$language = JFactory::getLanguage();`
`$language->load('', JPATH_ADMINISTRATOR, 'en-GB', true);`
`$language->load('', JPATH_ADMINISTRATOR, null, true);`
in the specific views, e.g. site\views\messagelist\view.html.php  line 108

st-anton:
I guess, this is needed in the administration pages of studies edit, media files edit and comments edit as well; not tested.

bcordis:
@st-anton can you add this to your language load scripts?

bcordis:
@st-anton This should be fixed now. Please re-test I updated paths to us are defines.

st-anton:
Test ok -> Closed

PR #477, opened 2015-02-09
Add Joomla standard language files to all Frontend admin pages

st-anton:
Solves #450


st-anton:
Added fix of missing plugin definitions

PR #499, opened 2015-03-04
Some phrases added (media and server)

st-anton:


st-anton:
Also added bugfix of #510

st-anton:
Is there any reason, not to include this changes?

Issue #529, opened 2015-03-28
Frontend biblestudy.css: a lot of leftovers ?


st-anton:
In media/css/biblestudy.css I found a lot of tags, tat cannto be found in the code like: bsmsdetailstable, seriestable, seriesstudytable.
Are they leftovers and can be deleted?

tomfuller2:
There should lots of those. If you can't find in the code they can be
deleted. Since changing the views over to taking their look from the site
template.

bcordis:
Yes they are created by the lss file if you find them let me know and I'll clean them up. We need to create some new ones anyway.

st-anton:
I collected information how the .css does work and got some ideas (and questions as well), how the situation may be improved:
- Currently the origin of the .css is the database table _bsms_style. Does it make sense to maintain the .css editor in the component? Each delevoper has a faster and more mighty editor (text highlighting, comparison...) than the buildt in editor.
- In the first attempt I was not aware, that the database is the master. So I first changed the .css file, then tried to adapt the database table. But saving overwrote the .css file and my work was lost. That was annoying.
- @bcordis you mention the lss (or .less?) file, the .css file is generated out of. Do you have a compiler this for? How is the result imported into the database table?
- I started to collect all the codes, that can be removed. And in fact, only a few will survive. @tomfuller2 you mentioned, that basically the look of the site template should be used. So the .css of the 9.0.0 version will be quite short.
- During migration the database table is used, somehow modified (\admin\helpers\dbhelper.php lines 342 to 350 seems to do so) and written to file. But as the old stuff is nearly completely removed, this is no more helpful to use the old items.
- As far as I remember, you planned to release a 8.0.9 version to fix the format of this table. Is this really needed / worth the work?
**Conclusion**: Does it make sense to simply drop the table _bsms_style, remove the css editor and deliver the css file as it is?

st-anton:
Created pull request #531 with reviewed database table _bsms_style as well as biblestudy.css

st-anton:
@bcordis: If you want to review the .less file, here are the changes:
Note 1: Several of them appears in biblestudy.less as well
- bsm_teachertable_list deleted, note 1
- bsm_teacherthumbnail_list deleted
- bsm_teachertable used in \com_biblestudy\site\views\teachers\tmpl\default_custom.php
- landingtable, landingrow, landingcell, landinglink used in \com_biblestudy\site\helpers\landing.php
[...]

PR #531, opened 2015-03-31
Remove unused item of css style; both in _bsms_style and generated biblestudy.css

st-anton:
This is a review of the css of issue #529 . Both files are synced to each other.
If the database table _bsms_style should not be maintained, the reviewed biblestudy.css may be used alone.


bcordis:
New css

PR #591, opened 2015-07-26
9.0.0 beta2

bcordis:
This will be the beta pull.


bcordis:
Completed test and moving into master as stable bate v2. Available for public bate

PR #641, opened 2016-03-20
9.0.1

bcordis:
Small fixes.


bcordis:
I have used phpstorm to inspect all the code and have found a few errors.

I corrected all i could and opened a few Issues to for @tomfuller2 to look into.

bcordis:
Ok now we have fixed the errors.  I still need to add some deletion to the next file set.

bcordis:
Moving to 9.0.1 do to update system. Will be releasing stable version once merged. Have a few media defaults to work on but close to being finished.

bcordis:
@tomfuller2 Look at the new system that now pulls over the defaults from Servers on Creation of new media. Work also if changing servers.

tomfuller2:
Looks good!

bcordis:
Delayed till Friday 15 2016

bcordis:
Close now as we have made it now progress on the stepper and how it will work with large data basses. 

I have a few more bugs to work out then we are going into testing again.

bcordis:
@tomfuller2 This is working now. I have a function that will need to be removed on release but allows you to update you db and run the install state every time.

I works right. Please debug.

PR #659, opened 2016-06-28
Working on issue with series breaking template if using landing page …

tomfuller2:
…limit. I believe I've fixed it.


bcordis:
You will need to move the comment down on its own line.

PR #679, opened 2016-07-31
Fix problem with teacher ID filtering

bcordis:
This one will fix the DB call with publish not being set within ().


bcordis:
@tomfuller2 Look at this. This fixes the problem on my website.

bcordis:
OK Just merged it as the change is tested and working.

PR #685, opened 2016-08-13
Fix problem 669

### LINKED ISSUE #669: Module doesn't pull in icon for default download icon
It only shows the default image.

bcordis:
Will fix #669


bcordis:
@tomfuller2 try this branch out.  It now passed the right info along to the module. Fixed some other errors with the checking if there is no filters set.

Issue #688, opened 2016-08-27
Brocken Model pulling up latest study.


bcordis:
Look like some how the model is not pulling the latest study. Will need to debug this problem.

bcordis:
I'll tackle this tomorrow as not sure whey this is happing on my website.  Hope it it not prevalent across the would.

bcordis:
@tomfuller2 Looks like the order of params is the problem need to redo to new one.

file mod_biblestudy.php
$admin        = JBSMParams::getAdmin();
$admin_params = $admin->params;

$admin_params->merge($template->params);
$admin_params->merge($params);
$params = $admin_params;

PR #691, opened 2016-09-02
Fix problem 688

bcordis:


bcordis:
Fix #688

bcordis:
redoing in new pull request

Issue #719, opened 2016-11-18
Error decoding JSON data: Syntax error 


jrockman13:
xposted: http://www.joomlabiblestudy.org/index.php?option=com_kunena&view=topic&catid=35&id=3734&Itemid=812#4484

jrockman13:
anyone had a chance to look at this?

jrockman13:
Update: posted new thread in correct forum: http://www.joomlabiblestudy.org/index.php?option=com_kunena&view=topic&catid=39&id=3735&Itemid=812#4488

bcordis:
Need a db backup to test bug. As right now can't reproduce the bug.

jrockman13:
Can I post it on Github or should I share a google drive link?

bcordis:
I would send a director linked on the ticket so I can look download the db. Not sure how big it is. Plus for privacy.

bcordis:
Please replace this file ASAP.

https://github.com/joomla/joomla-cms/blob/staging/libraries/vendor/joomla/registry/src/Format/Json.php

This will fix the problem for you. It is coming out next versions of Joomla

bcordis:
Also I have made some corrections to our scripts.

https://github.com/Joomla-Bible-Study/Joomla-Bible-Study/pull/720

I have tested it on your db and it corrects all the errors in your json params.

Let me know if you have trouble with this.

PR #720, opened 2016-12-08
Fixes some additional narrowing problems json errors. This fixes many …

bcordis:
…of the problems with decoding.


bcordis:
Fixes #719

PR #741, opened 2017-05-09
V9.0.15

bcordis:
Fix some install errors


bcordis:
After review commenting. =

PR #804, opened 2019-10-31
v9.2.1

bcordis:


bcordis:
Right now ignore the Travis CI checks they are not working yet. We will have to fix them later.

PR #811, opened 2020-05-15
Fixes to the Core

### LINKED ISSUE #810: Search Parameters not passing through to controller
Found and verified on https://www.nfsda.org and https://willows.org.za/

Need to fix search parameters.

bcordis:
This fixes the filters updating correctly on the page.
This also allows for the landing page to properly pass filters to the query to filter results. the filters are reflected on the sermons page.

Remove the old duration defined in the teacher's view.


bcordis:
Reworked the query build to foreach through building the query as we had redundant code. As menu settings will still override the filter set by the page. I need to look into a way to make them disappear if being overridden by a menu?

@tomfuller2  What is your thought of the rework.

bcordis:
Wow, I found a big speed improvement. Now it's loading a lot faster.

PR #818, opened 2020-09-07
9.2.4 work up

bcordis:
Work up for new version


bcordis:
Need review before so I can create Update list.

PR #824, opened 2020-12-13
Fixing Download URL problem making downloads not work right.

bcordis:
help speed up file size lookup and make sure the url is correctly formed.


bcordis:
Tested on Live site and fixes are good.

Issue #849, opened 2021-03-28
Modele Sermons Row Image span


bcordis:
Not following correct span selection. Selected Study but go teacher.

PR #850, opened 2021-03-28
#849_Fix_SermonsModele This fixes the problem with Model not fallowin…

bcordis:
…g the image span selection correctly. Found that selection for Study was rendering the same image.


bcordis:
#849 This fixes this problem.

bcordis:
Ready for Review.

PR #860, opened 2021-05-21
Jbsm media fsize

bcordis:
Fix int problem


bcordis:
Found problem with scripture text not rendering right and fixed it. Also found a problem with sermons rendering extra div that is not needed.

PR #863, opened 2021-06-01
Potential fix for the XSS problem

bcordis:


tomfuller2 (commented):
I haven't seen what the XSS is - but you believe this fixes it?

PR #902, opened 2022-09-27
Fix error loading the view

brianteeman:
in line 66 you are using _Route_ but you were not setting it up with a USE statement

You should check all the other views for the same

Signed-off-by: Brian Teeman <brian@teeman.net>


tomfuller2 (approved):
Thanks for the catch, Brian!

PR #920, opened 2023-01-02
use html5 instead of jwplayer for direct link; omit downloading pwplayer code

st-anton:


bcordis:
@st-anton Can you try and update this pull request now.

PR #966, opened 2024-10-03
New mod_proclaimicon for Admin View.

bcordis:


tomfuller2 (approved):
Let's test

tomfuller2 (approved):
APPROVED

bcordis:
@tomfuller2 Should we install this by default when we install?

tomfuller2:
Yes!

Tom Fuller, APR
Imagine Communications <http://www.ImagineCommunications.xyz>
971-832-9494

PR #993, opened 2025-09-07
Development j6

bcordis:
This branch is to fix all Joomla 6 compatibility errors.


bcordis:
Testing on ‎6.0.0-alpha3

PR #1212, opened 2026-05-01
fix(teacher): biography editors, social_links column, asset cleanup crash

bcordis:
## Summary

Three regressions on the teacher edit form, all surfaced from one repro on local + nfsda.org.

- **`Cwmassets::stripEmptyAssetRow` fatal**: `$table->getDatabase()` is `protected` on `Joomla\CMS\Table\Table` (J5+). Every save that hit the asset-strip path (teacher, message, series, location, mediafile, server, podcast, template, …) crashed with *Call to protected method* **after** `parent::store()` had already written the row. Switched to `Factory::getContainer()->get(DatabaseInterface::class)` like the rest of the file.
- **Biography tab editors not initialising**: the AJAX lazy-load added in 899cc9e8 ships editor HTML, but Joomla only loads `plg_editors_tinymce` JS and the `plg_editor_tinymce` script options when an editor field renders server-side. Initial page had no editor → injected markup had nothing to wire it up. Reverted to direct render; removed the now-dead `loadBiography` controller action.
- **`social_links` column missing on upgraded sites**: PR #1081 added a migration to add the column, but baf664f86 ("SQL migration cleanup") overwrote that file's contents with a bible-tables placeholder. Sites whose schema bump ran after that commit never got the column. Joomla silently drops unknown columns from `store()`, so only the first `facebook` URL persisted via the legacy `facebooklink` mapping — hence "only one is saving." Added `social_links` to the existing `requiredFixes` map in `proclaim.script.php`; `ensureSchemaReady()` repairs it on next preflight.

## Test plan

- [x] Teacher edit page renders Biography tab with both TinyMCE editors initialised on first paint
- [x] Save a teacher (existing or new) — no `Call to protected method` fatal in error log
- [x] Add 3+ social link rows of mixed platforms (e.g. facebook, instagram, youtube), save, reload — all rows persist
- [x] On a site missing the column, run a component update; verify `SHOW COLUMNS FROM <prefix>_bsms_teachers LIKE 'social_links'` returns a row after preflight
- [x] Sermon, series, location, mediafile, podcast, template saves still complete cleanly (same `stripEmptyAssetRow` path)


bcordis:
Pushed two related schema-form fixes (commit 2dad2181e):

- **Sermon `author`**: timing — `onContentAfterSave` (which builds the auto-schema) fires inside `parent::save()`, but `saveTeachers()` runs *after* `parent::save()`, so the junction was empty when `buildSermonSchema` queried it. Added `CwmschemaorgHelper::ensureSermonAuthor()` and call it inside `CwmmessageModel::saveTeachers()` — only fills `author` when missing, so a user-customized value on the Schema tab is preserved.
- **Teacher `worksFor`**: omission — `populateTeacher` (form-load auto-fill) didn't have the `worksFor` block that `buildTeacherSchema` and the Reset-Schema path both do. Field rendered empty, form posted back `['@type' => 'Organization', 'name' => '']`, and the preserve-complex-fields loop in `onSchemaPrepareSave` overrode the auto-derived value with the empty submission. Mirrored the block into `populateTeacher`.

Reset Schema remains the explicit refresh path for either schema.

### Updated test plan additions
- [ ] Create a new teacher with Org Name filled in → save → reopen, verify Schema tab `worksFor` is pre-filled
- [ ] Create a new sermon with at least one teacher in the subform → save → reopen, verify Schema tab `author` is pre-filled
- [ ] Customize `author` or `worksFor` on the Schema tab → save → verify the customization persists across subsequent saves (i.e. backfill is no-op when already set)

Issue #1273, opened 2026-06-27
Playlist sync with youtube


bcordis:
Given how YouTube works, we should consider adding support for YouTube playlist syncing and software-driven updates. This could be a template that updates this, as when Resi.io builds it, it misses the playlist, and there could be other things we should always sink when linked.

This is also a way to create YouTube events.

bcordis:
## #1273 — YouTube Playlist Sync + Write-back — Implementation Plan

> **Status:** Planned, not started. Saved 2026-06-27.

### Core concept (decided with maintainer)
A **Playlist is a first-class entity, distinct from a Series.** A playlist *may* back a
series (the "sermon series" case), but playlists also exist as cross-cutting groupings
(e.g. "All Church Services"). So we model Playlist as its own table + MVC, with an
**optional** `series_id` link — not a `playlist_id` column bolted onto `#__bsms_series`.

Proclaim becomes the **source of truth for default playlist settings**. Existing
YouTube videos currently miss those defaults; the feature both (a) reads playlist
videos into Proclaim and (b) writes Proclaim's defaults back to YouTube via OAuth,
including creating YouTube playlists/events.

### Scope (decided): BOTH read-sync AND OAuth write-back.

### What already exists (reuse)
- `CWMAddonYoutube::fetchPlaylistVideos()`, `fetchChannelPlaylists()` — fetch works (AJAX-only, not persisted)
- `CWMAddon::detectMetadata()`, `normalizeFilename()`, `buildEmbedUrl()` — media wiring
- `CwmcsvimportHelper::importRow()` — proven study+mediafile bulk-create + FK resolution pattern
- Task plugin `plugins/task/proclaim/` — `platformStatsTask` pattern (iterate servers → addon method → batch + quota)
- `CwmyoutubeQuota` — daily-budget guard; `CwmyoutubeLogHelper` — logging
- YouTube OAuth already wired in `youtube.xml` (`youtube_oauth` fieldset) — refresh-token grant (see memory youtube-apiclient-phpseclib)

### Phases (each ~1 PR, against `epic/1273-youtube-playlist-sync`)

**Phase 1 — Playlist entity (data model + MVC)**
- New table `#__bsms_playlists`: id, title, alias, description, `youtube_playlist_id`,
  `server_id` (FK servers), `series_id` (FK series, NULLABLE), `default_settings` (JSON:
  privacy, category, default description/tags template, location/teacher defaults),
[...]

PR #1284, opened 2026-06-27
feat(playlist): scheduled YouTube playlist sync task + conflict gate (#1273 phase 3)

bcordis:
## Phase 3 — Scheduled Sync + Conflict Gate (#1273)

Targets `epic/1273-youtube-playlist-sync` (stacked on phase 2).

Adds the `proclaim.playlistsync` scheduled task that keeps already-imported
playlists in step with the channel — **reusing the same `CwmplaylistSyncHelper`
engine as the "Import from YouTube" toolbar action**, so the manual and
scheduled paths can never drift.

### Gate logic (as requested)
- **Initial import stays deliberate.** The task defaults to `discover_new = 0`:
  it only refreshes playlists a human already imported and never creates a
  channel's playlists unattended. New / opted-out playlists are counted as
  *skipped* and reported ("run Import from YouTube to add them").
- **Neither side silently overwrites the other.** A conflict gate re-applies the
  YouTube title to a playlist only when it *changed* **and** the local row was
  not edited since the last sync (`modified <= last_sync`). When both diverged,
  the local value is kept and the divergence is logged as a conflict. "Only
  write if changed" avoids needless writes. (True bidirectional resolution comes
  with phase-5 write-back; until then read-sync prefers the local edit.)

### Changes
- `CwmplaylistSyncHelper::import()` gains a `$discoverNew` flag + `playlistsSkipped`
  / `conflicts` stats; `importChannelPlaylists()` gains the gate + `isLocallyEdited()`.
- New task: `TASKS_MAP` entry, `playlistSyncTask()`, `forms/playlistsync.xml`
  (`server_id` + `discover_new` switch), language strings.
- Toolbar import now surfaces conflicts as warnings too.

### Verification (live, j6-dev server 6)
- Task path (`discover_new=0`): refreshes 22 existing playlists, skips 0.
- Conflict gate: a local edit made *after* last sync is **preserved and reported**.
- One-sided remote change (stale local, no edit since sync): YouTube title
  correctly **applied**.
- `composer test:unit`: **696 tests / 1403 assertions green.**

### Possible follow-up (raised by maintainer)
"Sync on save of the media/message" — reconcile a single YouTube media file into
its playlist immediately when it's saved, instead of waiting for the task. Would
be a small hook in the mediafile model save calling the same engine. Left out of
this PR pending confirmation.


bcordis:
### Added: phase 3.5 — on-save playlist linking (commit cc5c03ca1)

The "sync on save of the media/message" follow-up is now in this PR (confirmed by maintainer).

When a YouTube media file is saved, it's attached to the playlist(s) that already list its video **immediately**, instead of waiting for the scheduled task. It's **local-only** (no API call) — it backfills the junction rows a prior import left unmatched (`mediafile_id NULL`) because the media didn't exist yet.

- `CwmplaylistSyncHelper::linkMediafileToPlaylists()` (fill unmatched rows for the video; clear stale links if the URL changed) + `unlinkMediafile()` (null `mediafile_id` on delete, keep the row).
- Hooked into `CwmmediafileTable::store()` / `delete()` — the universal persistence point for the admin form, the message wizard, and the API.
- Never throws: a media save/delete can't fail on playlist bookkeeping; safe no-op for non-YouTube media and on sites without the playlist tables.

Verified live on j6-dev through the real Table hook: saving re-links the previously-unmatched row; delete unlinks while preserving the row. 696 unit tests green.

PR #1304, opened 2026-07-17
fix(release): repair pkg assembly + clean-install install-script bugs

### LINKED ISSUE #2: Some files not deleted which kills site
/site/helpers/image.php and helper.php not deleted on upgrade. image.php kills the biblestudy listing and must be renamed.

bcordis:
## Context

While building a release-validation harness (separate PR), building and clean-installing `pkg_proclaim` end-to-end via the **Extensions web installer** surfaced four pre-existing, **release-blocking** defects. Clean-install e2e had never actually been exercised (see the `pkg_proclaim Issues` note), so these went unnoticed.

Each was verified fixed by a clean install on a pristine Joomla 6.1 site.

## Packaging — `pkg_proclaim` could not be assembled at all

- `cwm-build.config.json`'s `package.includes` still referenced a `build/build-package.php` sub-build script that **no longer exists** in either sibling submodule (they migrated to their own `cwm-build` / `cwm-package`). `cwm-package` aborted before assembling. Switched both entries from `subBuild` → `prebuilt`, globbing each submodule's own `build/dist` output.
- Added **`build/build-subpackages.sh`** (builds `lib_cwmscripture` via `cwm-build` and `content/scripturelinks` via `cwm-package`, fresh) and **`build/build-package.sh`** (sub-packages → `cwm-build` → `cwm-package`) so the `prebuilt` artifacts exist.

## Clean install (`proclaim.script.php`)

1. **preflight** checked `libraries/lib_cwmscripture/src`, but the library manifest declares `<libraryname>cwmscripture</libraryname>` → it installs to `libraries/cwmscripture/src`. The namespace never registered and every fresh install aborted with *"requires lib_cwmscripture to be installed first."* Now checks the real path (old name kept as fallback).
2. **postflight** `require_once`'d `ScriptureReference.php`, which **moved to lib_cwmscripture in 10.3.0**. A bare `require_once` of the missing file is a fatal the surrounding `catch` can't trap. Guarded with `is_file`.
3. The block of **legacy data migrations** (`migrate*` / `dropLegacy*` / `fix*Legacy*`) ran under `install || update`. On a fresh install there is nothing to migrate; running them against freshly-seeded defaults produced a spurious *"1 podcast link(s) could not be matched to a menu item"* warning (and, before fix #2, a fatal). Gated the whole block to `$type === 'update'`; guided tours, `helpURL`, and `ensurePrimaryKeys` stay on both.

## Verification

Clean install via **Extensions → Install → Install from Folder** on a pristine J6.1 site: **"Installation of the package was successful"** with no spurious warnings; all 27 `#__bsms_*` tables created; every 10.3.3 migration column/table/index present; `#__schemas` at `10.3.3`.

### DIFF
--- a/build/build-package.sh
+++ b/build/build-package.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# Build the full installable pkg_proclaim-<version>.zip end to end:
+#   1. build the sibling sub-packages (lib_cwmscripture, scripturelinks) fresh
+#   2. build com_proclaim's assets + component zip  (cwm-build)
+#   3. assemble the package wrapper                 (cwm-package)
+#
+# This is the single command the release-test harness and the release runbook
+# use to produce the artifact users install. Non-interactive.
+#
+# Usage: build/build-package.sh <version>
+#
+# @since __DEPLOY_VERSION__
+
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT"
+
+BIN="libraries/vendor/bin"
+VERSION="${1:-}"
+
+if [ -z "$VERSION" ]; then
+    echo "ERROR: version required — usage: build/build-package.sh <version>" >&2
+    exit 1
+fi
+
+echo "-- [pkg 1/3] build sibling sub-packages"
+bash build/build-subpackages.sh
+
+echo "-- [pkg 2/3] build com_proclaim assets + component (${VERSION})"
+CWM_NONINTERACTIVE=1 "$BIN/cwm-build" --version "$VERSION"
+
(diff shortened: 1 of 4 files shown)


bcordis:
### Impact note: this unbreaks fresh installs across the whole 10.3.x line

Verified on a pristine Joomla 6.1 site: the **released 10.3.2** package fails a fresh install — the component installs, then postflight dies with `Failed opening required .../com_proclaim/src/Helper/ScriptureReference.php` (bug #3 here). The same class of bugs (preflight lib path, this fatal `require_once`) is present in 10.3.0–10.3.2, so **no shipped 10.3.x version fresh-installs cleanly**. Real sites only reach 10.3.x by upgrading from ≤10.2.x, which is why this went unnoticed.

This PR is therefore not cosmetic — it restores the clean-install path for the entire 10.3 line, first shipping in 10.3.3. Tracking issue for the shipped-versions state: (linked below).

PR #1367, opened 2026-07-29
fix(assets): register the lib_cwmscripture asset registry file

### LINKED ISSUE #1366: Scripture version switcher is non-functional: com_proclaim never registers the lib_cwmscripture asset registry file
## Symptom

The scripture version switcher does not work on Proclaim site views.

## Root cause

`com_proclaim` enqueues the switcher assets by name but never registers the registry file that defines those names.

`site/src/Helper/Cwmshowscripture.php:366-367` (inside `renderVersionSwitcher()`):

```php
$wa->useScript('lib_cwmscripture.scripture-switcher');
$wa->useStyle('lib_cwmscripture.scripture-switcher-css');
```

and again at `site/src/Helper/Cwmshowscripture.php:634-635` (inside `renderUnavailableNotice()`).

Those asset names are defined in `media/lib_cwmscripture/joomla.asset.json`, which ships with the library. Joomla's `WebAssetRegistry` only auto-loads the registry files for core, the active template, and the **active extension** — it does not discover a library's file on its own. It has to be registered explicitly with `addExtensionRegistryFile('lib_cwmscripture')`.

The only such call anywhere in this repository is in the **plugin**:

- `plugins/content/scripturelinks/src/Extension/ScriptureLinks.php:1134` — `$wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');`

Nothing in `site/` or `administrator/` registers it. So on a sermon view rendered by the component, the asset names are unknown to the WebAssetManager when `useScript()` runs. `WebAssetRegistry::get()` throws `UnknownAssetException` for an unregistered name, so the result is either a bubbled exception or — where it is swallowed by one of the surrounding `try/catch (\Exception)` blocks in this helper — switcher markup rendered with no JS or CSS behind it, i.e. an inert control.

This is consistent with the switcher having worked previously: the assets used to live under `media/com_proclaim/` (still referenced in the cleanup list at `proclaim.script.php:234` and `:245`) and were moved into lib_cwmscripture. Under the old paths the component *was* the active extension, so registration was automatic.

## Reproduction

1. Enable **Allow version switch** (`allow_version_switch = 1`) in Proclaim's scripture params — `renderVersionSwitcher()` is gated on this.
2. View a sermon on the site with a scripture reference attached, on a page where the `content - scripturelinks` plugin has not fired.
3. The switcher renders but does not respond; the switcher JS/CSS are absent from the page source.

## Fix

Register the library's registry file before enqueueing, in both call sites:

```php
$wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
```

`addRegistryFile()` is idempotent, so calling it on every render is safe.

Companion hardening issue in the library, so consumers cannot forget this step: Joomla-Bible-Study/lib_cwmscripture#13

Per Joomla-Bible-Study/lib_cwmscripture#10, fixes to CWM-owned code land in the owning repo — the component-side registration is this repo's, the library-side helper is theirs.

bcordis:
## Problem

The scripture version switcher is non-functional on site views, and the same defect kills the scripture tooltips.

`com_proclaim` enqueues `lib_cwmscripture.*` assets by name, but never registers `media/lib_cwmscripture/joomla.asset.json` — the file that defines those names. Joomla's `WebAssetRegistry` auto-loads registry files for core, the active template and the active **extension** only; a library's file is never discovered on its own. `WebAssetRegistry::get()` then throws `UnknownAssetException` for the unregistered name, so the result is either a bubbled exception or — where one of the surrounding `try/catch (\Exception)` blocks swallows it — markup rendered with no JS or CSS behind it.

The only correct call site in the repo was in the plugin (`plugins/content/scripturelinks/src/Extension/ScriptureLinks.php:1134`). That meant the assets resolved only on pages where `content - scripturelinks` had already run; on a sermon view rendered by the component, they did not.

This is consistent with the switcher having worked before: these assets used to live under `media/com_proclaim/` (still in the cleanup list at `proclaim.script.php:234`/`:245`) and were moved into lib_cwmscripture. Under the old paths the component *was* the active extension, so registration was automatic.

## Change

Adds the idempotent `$wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');` ahead of all five affected enqueue sites — 5 added lines, no other changes:

| Location | Assets |
|---|---|
| `Cwmshowscripture::renderTextPassage()` | `scripture-text` |
| `Cwmshowscripture::renderVersionSwitcher()` | switcher JS + CSS |
| `Cwmshowscripture::renderUnavailableNotice()` | switcher JS, `scripture-text` |
| `Cwmsermon\HtmlView` | tooltip JS + CSS |
| `Cwmsermons\HtmlView` | tooltip JS + CSS |

Registering directly (rather than calling the library's new helper) keeps this fix independent of the bundled library version, so it works against the currently shipped lib_cwmscripture.

## Testing

`php -l` clean on all three files. Grep confirms every `useScript('lib_cwmscripture`/`useStyle('lib_cwmscripture` call in `site/` is now preceded by a registration.

Worth verifying on a live sermon view with **Allow version switch** (`allow_version_switch = 1`) enabled, since `renderVersionSwitcher()` is gated on that param.

## Related

Library-side hardening: Joomla-Bible-Study/lib_cwmscripture#14 makes `ScriptureRenderer` self-sufficient and adds `loadSwitcherAssets()`, so consumers cannot forget this step. These call sites can migrate to it once the bundled library is bumped past that release.

Closes #1366

### DIFF
--- a/site/src/Helper/Cwmshowscripture.php
+++ b/site/src/Helper/Cwmshowscripture.php
@@ -217,6 +217,7 @@ class Cwmshowscripture
     public function renderTextPassage(BiblePassageResult $result, int $choice, Registry $params, string $switcherHtml = ''): string
     {
         $wa = Factory::getApplication()->getDocument()->getWebAssetManager();
+        $wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
         $wa->useStyle('lib_cwmscripture.scripture-text');
 
         $copyrightHtml = '';
@@ -363,6 +364,7 @@ class Cwmshowscripture
     {
         $app = Factory::getApplication();
         $wa  = $app->getDocument()->getWebAssetManager();
+        $wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
         $wa->useScript('lib_cwmscripture.scripture-switcher');
         $wa->useStyle('lib_cwmscripture.scripture-switcher-css');
 
@@ -631,6 +633,7 @@ class Cwmshowscripture
     private function renderUnavailableNotice(object $row, string $reference, string $version): string
     {
         $wa = Factory::getApplication()->getDocument()->getWebAssetManager();
+        $wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
         $wa->useScript('lib_cwmscripture.scripture-switcher');
         $wa->useStyle('lib_cwmscripture.scripture-text');
 
--- a/site/src/View/Cwmsermon/HtmlView.php
+++ b/site/src/View/Cwmsermon/HtmlView.php
@@ -500,6 +500,7 @@ class HtmlView extends BaseHtmlView
         // Load scripture tooltip assets (per-element controlled; JS is a no-op
         // if no elements have show_tooltip enabled). Skip in print mode.
         if (empty($this->print)) {
+            $wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
             $wa->useScript('lib_cwmscripture.scripture-tooltip');
             $wa->useStyle('lib_cwmscripture.scripture-tooltip');
 
--- a/site/src/View/Cwmsermons/HtmlView.php
+++ b/site/src/View/Cwmsermons/HtmlView.php
@@ -301,6 +301,7 @@ class HtmlView extends BaseHtmlView
 
         // Load scripture tooltip assets (per-element controlled; JS is a no-op
         // if no elements have show_tooltip enabled)
+        $wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
         $wa->useScript('lib_cwmscripture.scripture-tooltip');
         $wa->useStyle('lib_cwmscripture.scripture-tooltip');
 


bcordis:
Closing: the switcher is fixed by #1370, and the change here would not have fixed it.

The premise — that `com_proclaim` never registers `lib_cwmscripture`'s asset registry file — does not hold. `admin/api.php:111` does it unconditionally:

```php
$wa->getRegistry()->addExtensionRegistryFile('lib_cwmscripture');
```

and that file is `require_once`'d by every entry point that renders these assets: `site/src/Dispatcher/Dispatcher.php:16-20`, `admin/src/Dispatcher/Dispatcher.php`, and `modules/site/mod_proclaim/src/Dispatcher/Dispatcher.php:46-47`. All nine asset names com_proclaim requests resolve against the library's `joomla.asset.json`, so there was no `UnknownAssetException` to prevent.

The actual cause was a script-options key mismatch from the 10.3.0 library split: `scripture-switcher.es6.js` reads `cwmscripture.options` with no fallback, while the component published `com_proclaim.scripture`, so `ajaxBaseUrl` was `''` and every switch hit an early return that only logs under Site Debug. Full analysis in the correction comment on #1366; fix in #1370.

The calls added here are harmless — `addRegistryFile()` is `is_file()`-guarded and dedupes on path — but they are no-ops, and leaving them in would suggest the registry was the problem, which would mislead the next person to read this code.

PR #1395, opened 2026-08-01
fix(podcast): four RSS/Apple spec defects in the generated feed

### LINKED ISSUE #1391: Podcast feed: four spec-compliance defects in Cwmpodcast.php
Auditing a production `audio.xml` against the RSS 2.0 / Apple Podcasts specs surfaced four defects in feed generation. All four are in `site/src/Helper/Cwmpodcast.php` and all four affect every site running the component, not just the one audited. Filing together since they're all in the same file and three are one-line fixes; happy to split if that's easier to triage.

---

## 1. `itunes:duration` is not zero-padded

`getEpisodeDuration()` concatenates the three stored params verbatim:

```php
$hours   = $episode->params->get('media_hours', '00');
$minutes = $episode->params->get('media_minutes', '00');
$seconds = $episode->params->get('media_seconds', '00');

if ($hours !== '00' || $minutes !== '00' || $seconds !== '00') {
    return $hours . ':' . $minutes . ':' . $seconds;
}
```

Nothing guarantees the stored values are two digits. `fixMediaDurations()` pads on write (`str_pad(..., 2, '0', STR_PAD_LEFT)`), but records entered manually through the media edit form do not go through that path, so a single-digit seconds value is stored bare and emitted bare.

**Result:** `<itunes:duration>00:27:7</itunes:duration>` — invalid. Apple's validator rejects it; several clients fall back to reading duration from the file, and some display nothing.

In the audit sample this hit 4 of 30 episodes (~13%), all with single-digit seconds.

**Fix** — the correct pattern already exists elsewhere in this same file (line ~1626):

```php
return sprintf('%02d:%02d:%02d', (int) $hours, (int) $minutes, (int) $seconds);
```

Note the guard above it is also a strict string comparison against `'00'`, so a value stored as `'0'` passes the guard and reaches output. Casting to int in the sprintf handles that too, but the guard is worth revisiting — `if ((int) $hours || (int) $minutes || (int) $seconds)` expresses the intent better.

There's already a `tests/unit/Site/Helper/CwmpodcastDurationTest.php` — this looks like a straightforward case to add.

---

## 2. `<enclosure type>` is emitted unvalidated, and the fallback is not a registered MIME type

In `getEnclosureXml()`:

```php
$mimeType = $episode->params->get('mime_type');
// ...
} elseif ($docmanId > 1) {
    $type = $mimeType;                          // no fallback at all
} else {
    $type = $mimeType ?: 'audio/mpeg3';         // not a registered type
}
```

Three problems:

- **`audio/mpeg3` is not an IANA-registered media type.** The registered type for MP3 is `audio/mpeg`. (`audio/mpeg3` and `audio/mp3` are both common-but-wrong values that circulate widely.)
- **The DOCman branch has no fallback**, so an empty `mime_type` emits `type=""`.
- **Whatever the admin typed goes straight into the XML.** In the audited feed every episode carried `type="audio/mp3"` — also unregistered — because that string was stored on the media records and the component passed it through unchanged.

Worse, because the type is stored rather than derived, it can contradict the actual file. The audited feed had two `.m4a` files declared as `audio/mp3`. Nothing in the component detects that mismatch.

**Fix:** derive from the file extension with a stored override, rather than trusting the stored value outright. The class already has exactly this pattern for transcripts (~line 493):

```php
$mimeMap = [
    'vtt'  => 'text/vtt',
    'srt'  => 'application/x-subrip',
    'txt'  => 'text/plain',
    'json' => 'application/json',
];
// ...
$ext  = strtolower(pathinfo(parse_url($src, PHP_URL_PATH) ?: $src, PATHINFO_EXTENSION));
$type = $mimeMap[$ext] ?? 'text/vtt';
```

Generalizing that map to cover media (`mp3 => audio/mpeg`, `m4a => audio/x-m4a`, `mp4 => video/mp4`, `m4v => video/x-m4v`, `aac => audio/aac`, `ogg => audio/ogg`, `opus => audio/opus`, `wav => audio/wav`, `webm => video/webm`) and using the stored `mime_type` only when the extension is unrecognized would fix the pass-through, the bad default, and the empty-string case in one change.

The existing feed validator (`JBS_PDC_VALIDATE_MEDIA_NO_MIME`) already warns on a *missing* mime_type. Extending it to also warn when the stored type contradicts the file extension would catch the `.m4a`-as-mp3 class of problem in the admin UI.

---

## 3. `<guid>` is missing `isPermaLink="false"`

```php
return '
        <enclosure url="' . $enclosureUrl . '" length="' . $size . '" type="' . $type . '" />
        <guid>' . $guid . '</guid>';
```

Per RSS 2.0, `isPermaLink` defaults to **true** when the attribute is absent — meaning readers are entitled to treat the guid as the URL of the episode's web page.

`CwmpodcastTrackHelper::resolveGuid()` returns a media file URL, not an HTML page. Some aggregators will attempt to fetch and render it as the episode permalink.

This matters more since #1299 made GUIDs permanent and decoupled them from the enclosure URL — the guid is now explicitly an identity token, so it should say so:

```php
<guid isPermaLink="false">
```

---

## 4. Protocol defaults to `http://`, and root-relative paths produce a malformed URL

```php
$protocol = $params->get('protocol', 'http://');
```

`resolveUrl()` prepends this to any value that isn't already absolute, and `$protocol` is threaded through channel `<link>`, `atom:link rel="self"`, `itunes:image`, `podcast:image`, `podcast:person img`, chapters, transcripts, and enclosure URLs.

On any site that hasn't explicitly changed the setting, **every URL in the feed is `http://`**. On an HTTPS site that 301s (which is essentially all of them now) each of those pays a redirect, and mixed-content warnings are possible for the artwork.

Defaulting to `https://` seems right in 2026. Deriving from `Uri::root()` / Joomla's `live_site` would be better still — it would track the site's actual configuration instead of a separately-maintained setting that silently drifts.

**Separately**, `resolveUrl()` has an unhandled input class:

```php
private function resolveUrl(string $value, string $protocol): string
{
    if (preg_match('#^[a-z][a-z0-9+\-.]*://#i', $value)) {
        return rtrim($value, '/');
    }

    if (ctype_digit($value)) {
        return Route::link('site', 'index.php?Itemid=' . $value);
    }

    return $protocol . $value;   // <-- root-relative input lands here
}
```

A root-relative value — e.g. a podcast link stored as `/resources/sermons.html`, which the admin field accepts — matches neither branch and falls through to string concatenation, yielding:

```
http:///resources/sermons.html
```

Triple slash, empty authority. Worth a `str_starts_with($value, '/')` branch that joins against the site root rather than concatenating onto the bare protocol.

---

## Environment

- Observed against a production feed generated by Proclaim; code references above are against current `HEAD`.
- The audited site is on an older release, so #4's relative-path symptom may present differently there — but the `resolveUrl()` fall-through is reproducible from the current source as written.
- Defects 1–3 are confirmed present in `HEAD`.

bcordis:
Fixes #1391

All four audited defects, plus one instance of the same bug the audit didn't reach.

## 1. `itunes:duration` zero-padding

`getEpisodeDuration()` concatenated the three stored params verbatim. `fixMediaDurations()` is the only thing that pads on write, and records entered through the media edit form never go through it — so `00:27:7` reached the feed. Now `sprintf('%02d:%02d:%02d', …)`.

The guard above it was a strict `!== '00'` comparison, so a value stored as `'0'` passed it and produced a bogus duration; it now compares numerically.

**Also fixed, not in the issue:** both feed validators carried that same string comparison in their missing-duration check, so a `'0'` duration was never reported to the admin. They compare numerically too — otherwise the validator would stay quiet about exactly the records that used to emit bad output.

## 2. Enclosure `type`

Three problems in one line: the stored `mime_type` went straight into the XML, the fallback `audio/mpeg3` is not IANA-registered, and the DOCman branch had no fallback at all so an empty type emitted `type=""`.

The type is now derived from the file extension through `Cwmmime::fromExtension()`, with the stored value honoured only when the extension yields nothing, falling back to `audio/mpeg`. That fixes the pass-through, the bad default, the empty attribute, and the `.m4a`-declared-as-MP3 mismatch the audit found. `podcast:alternateEnclosure` follows the same rule — it had the same `audio/mpeg` blanket default.

I used the existing `Cwmmime` helper rather than generalising the transcript map as the issue suggested: `Cwmmime` already exists precisely to stop per-file extension maps proliferating, and it already covers every media extension listed in the issue. Note it maps `m4a` to `audio/mp4` (registered) rather than the `audio/x-m4a` the issue proposed.

## 3. `<guid isPermaLink="false">`

One line. Since #1299 froze guids and decoupled them from the enclosure URL, the guid is an identity token, and RSS 2.0's default of `true` invites readers to fetch it as a web page.

## 4. Protocol default and `resolveUrl()`

The three PHP fallbacks now default to `https://`, matching what `admin/forms/template.xml` has defaulted to for some time — the PHP values are what apply to templates saved before that field existed. All three moved together deliberately: `CwmpodcastController` reconstructs the URL the enclosure declares for the tracking redirect, so a split default would send subscribers to a different origin than the feed advertises.

`resolveUrl()` gained a root-relative branch. A value like `/resources/sermons.html` matched neither existing branch and fell through to concatenation, producing `https:///resources/sermons.html` — triple slash, empty authority. It now joins against `Uri::root()`.

**I did not derive the protocol from `Uri::root()`** as the issue floated. The field offers `ftp://`, `icy://` and `rtmpt://` for streaming servers, so deriving it would break sites whose media lives on a different scheme than the site itself. Flipping the default gets the benefit without that risk.

## Testing

`tests/unit/Site/Helper/CwmpodcastFeedSpecTest.php` — 26 tests covering all four defects: duration padding across padded / single-digit / integer / all-zero inputs, MIME derivation including the contradicting-stored-value and unknown-extension cases, an explicit guard that `audio/mpeg3` can never be emitted again, the guid attribute, and each `resolveUrl()` input class.

22 of the 26 fail against the pre-fix source. Suites: Site Helper 115 pass, Admin Helper 521, Site Controller 4, Site Model 22. php-cs-fixer clean.

## Not done here

The issue also suggests extending the validator to warn when a stored `mime_type` contradicts the file extension. That needs a new language key across six locales, so it's filed separately rather than bundled — the feed output itself is no longer affected by the mismatch, since the type is derived rather than trusted.


bcordis:
Follow-up commit after a review question about the field defaults — the first pass only fixed the render path.

**The stored value was the root cause.** The duration fields are `type="text"` with `default="00"` and no `filter`, so a hand-typed `7` persists as `7`. Every automated writer pads (`str_pad` across eight call sites); manual entry was the only path that didn't. `CwmmediafileModel::save()` now normalizes the three parts alongside the icon-class and filename normalization already there, so the malformed value never reaches storage.

**And it was poisoning more than the feed.** Five "is the duration set?" checks compared against the string `'00'`, so a record holding `'0'` read as *set*:

| site | consequence |
|---|---|
| `fixMediaDurations()` | skipped the records it exists to repair |
| `getMediaNeedingDuration()` | omitted them from the report |
| `CWMAddon::needsDetection()` | skipped auto-detection |
| `fixMediaMetadata()` | reported "already set", did nothing |
| both feed validators | never warned (in the first commit) |

All compare numerically now. The metadata-fix success message also built its display string with `%s:%s:%s`, echoing the unpadded value back at the admin — now `%02d`.

Tests: 34 in `CwmpodcastFeedSpecTest` (8 new). Site Helper 123, Admin Helper 521, Admin Model 16, Addon integration 15 — all pass.

Issue #1609, opened 2026-08-06
Proclaim config's "Analytics retention days" setting is dead — nothing reads it


bcordis:
Found while fixing #1571 / #1608.

`analytics_retention_days` is defined in `admin/forms/admin.xml:309` (with `default="90" min="7" max="3650" filter="int"`) and rendered in `admin/tmpl/cwmadmin/edit.php:876`. 

**No PHP anywhere reads it.**

```
$ grep -rn "analytics_retention_days" . --include="*.php"
admin/tmpl/cwmadmin/edit.php:876:  <?php echo $this->form->renderField('analytics_retention_days', 'params'); ?>
```

The only consumer of a retention window is the task scheduler plugin, which reads its **own** `retention_days` from the task params (`plugins/task/proclaim/forms/analytics.xml`), not this one.

## Why it matters

An admin who sets retention to 365 in Proclaim → Settings gets no protection at all — their events are still purged on the task's schedule at whatever the task says. It is a user-facing data-retention control that silently does nothing, which is worse than not offering it, particularly since the setting governs irreversible deletion.

## Options

1. Have the task fall back to the component setting when its own param is absent (keeps both, component acts as the default).
2. Make the component setting authoritative and drop the per-task field.
3. Remove the dead component field and document that retention lives on the task.

(1) is probably the least surprising, but it is a behaviour choice worth deciding deliberately rather than defaulting into. Whichever is chosen, the floor added in #1608 applies at the point of deletion regardless.

bcordis:
Fixed in #1645.

Option 3: the component field is removed, and retention lives on the purge task.

Option 1 (task falls back to the component setting) was considered and rejected. `enable_purge` defaults to **off** because analytics history is kept indefinitely unless an administrator deliberately opts in; a component field defaulting to 90 days would reintroduce an implied retention limit that contradicts that. As this issue notes, a control that silently does nothing is worse than not offering one — particularly for irreversible deletion — and the fix is to stop offering it, not to wire it up to a default nobody chose.

Removed: the field in `admin/forms/admin.xml`, its `renderField()` call, and the now-orphaned `JBS_ANA_RETENTION_LABEL` / `_DESC` strings in **all seven languages**. The task's own description now states it is the only retention setting, so anyone looking for the removed one is pointed at where retention actually happens.

The floor added in #1608 still applies at the point of deletion, unchanged.

[...]

PR #1610, opened 2026-08-06
fix: CwmanalyticsHelper w-domain corruption, rollup atomicity, and purge safety

### LINKED ISSUE #1571: CwmanalyticsHelper: ltrim() www-strip bug corrupts referrer domains, non-atomic rollup double-counts on retry, unique key never fires (NULL columns), country_code never populated
Found during a Helper-folder-wide bug-hunting review, tier 3.

## 1. `ltrim($host, 'www.')` strips a character mask, not the "www." prefix

`admin/src/Helper/CwmanalyticsHelper.php:160,243,247`. PHP's `ltrim($str, $chars)` treats the second argument as a character set to strip, not a literal prefix. `ltrim('worship.example.org', 'www.')` strips only the leading `w` (next char `o` isn't in the mask) → `"orship.example.org"`. This corrupted value is stored into `referrer_domain` with no further validation. Any domain starting with `w` (worship, watch, webex, wesleyan, wordoflife) or containing an internal dot early triggers this — silent, permanent corruption of a stored analytics field, corrupting the "Top Referrer Domains" report.

**Fix direction:** Use a literal prefix strip: `str_starts_with($host, 'www.') ? substr($host, 4) : $host`.

## 2. `rollupAndPurge()` rollup+purge not atomic — retry after partial failure double-counts monthly totals

`CwmanalyticsHelper.php:385-444`. Runs as a scheduled task. If the rollup INSERT succeeds but the task is killed (execution-time limit, scheduler timeout, OOM, deploy restart) before the subsequent purge DELETE runs, the raw events are never purged. The next scheduled run's retention cutoff still includes those same rows, so they get aggregated again — `ON DUPLICATE KEY UPDATE count = count + VALUES(count)` adds to the existing row, doubling the count. Never self-correcting.

**Fix direction:** Wrap the rollup INSERT and purge DELETE in `$db->transactionStart()`/`transactionCommit()` (rollback on exception).

## 3. `ON DUPLICATE KEY UPDATE` never fires because the unique key includes always-NULL columns — rollup never actually consolidates

`CwmanalyticsHelper.php:396-429`. The `uq_aggregate` unique key on `#__bsms_analytics_monthly` (`admin/sql/install.mysql.utf8.sql:1226-1230`) includes nullable `series_id`/`study_id`/`media_id`/`location_id`/`country_code` — and `country_code` is always NULL (see finding 4). MySQL treats `NULL != NULL` in unique indexes, so this composite key never registers as a duplicate — every scheduled run inserts a brand-new row instead of consolidating into the prior month's row, defeating the table's stated purpose of bounded compact history. This exact defect is independently acknowledged elsewhere: `CwmanalyticsModel::seedFromLegacy()`'s docblock states verbatim the same root cause and applies a delete-then-reinsert workaround there — never applied to `rollupAndPurge()`.

**Fix direction:** Give the nullable dimension columns non-NULL sentinel defaults (e.g. `0`), or replace the blind `INSERT...ON DUPLICATE KEY UPDATE` with the same explicit delete-then-reinsert pattern `seedFromLegacy()` already uses.

## 4. `country_code` is claimed to be classified but is never populated — GeoIP breakdown is permanently empty

`CwmanalyticsHelper.php:80-220`. The class docblock states `logEvent()` "Classifies UA, referrer, and GeoIP at log-time." `logEvent()`'s INSERT column list never includes `country_code` — no GeoIP lookup exists anywhere in the codebase (confirmed via repo-wide grep). Every row is inserted with `country_code = NULL`. `CwmanalyticsModel::getBreakdown('country_code', ...)` ("Views by Country" dashboard widget) is permanently empty, contradicting the feature's own documentation.

**Fix direction:** Either wire up an actual GeoIP resolution step (discarding the raw IP per the existing "IP never stored" schema comment) and add `country_code` to the INSERT, or update the docblock and hide the country breakdown UI if GeoIP was deliberately deferred.

### LINKED ISSUE #1608: CwmanalyticsHelper: a blank retention field silently deletes the entire analytics history; enable_purge is ignored
Found while fixing #1571. Two ways the analytics purge destroys data that is still in active use.

## 1. A blank retention field wipes the whole events table

`plugins/task/proclaim/src/Extension/Proclaim.php:367`

```php
$retentionDays = (int) ($params->retention_days ?? 90);
```

`?? 90` only substitutes on **NULL**. A blank number field in the Scheduler UI saves as `""`, and `(int) "" === 0`. Verified:

| stored value | retentionDays | cutoff | effect |
|---|---|---|---|
| absent / null | 90 | now −90d | correct |
| `""` (blank field) | 0 | **now** | deletes every raw event |
| `"0"` | 0 | **now** | deletes every raw event |
| `"abc"` | 0 | **now** | deletes every raw event |
| `"-5"` | −5 | **now +5d** | deletes every raw event |

The purge is `DELETE FROM #__bsms_analytics_events WHERE created < :cutoff`. With the cutoff at (or after) `now`, that matches the entire table. Silent and permanent — the raw events are the only per-event record.

`min="7"` on the form field is a client-side rendering hint, not server-side validation. **Adding `filter="int"` does not fix this** — it converts `""` into exactly the lethal `0`. (Note `admin.xml`s copy of this field *does* have `filter="int"`, which is why it looks defended.)

This is reachable from ordinary use, not tampering: opening the task to retype the value and saving it blank is enough. Bad values can also arrive from a restored backup — `Cwmbackup` does `DELETE FROM #__scheduler_tasks` then re-INSERTs — a future migration, or a direct DB edit.

## 2. `enable_purge` is computed and then ignored

`Proclaim.php:366` reads `enable_purge`, but `rollupAndPurge()` took no such parameter and always deleted. An admin who explicitly set **Purge: No** still lost their raw events. `enable_rollup=0` + `enable_purge=1` was also honoured literally, deleting events that had never been aggregated anywhere.

## Fix (in the #1571 PR)

- `CwmanalyticsHelper::isPurgeSafe()` — pure, public, testable without touching the DB.
- Floor enforced **at the point of deletion**, so every caller is covered rather than enumerating entry points.
- An out-of-range window aborts **both** halves: `$retentionDays` *is* the cutoff, so a value of 0 would also have the rollup summarise todays events while the raw rows survive, inflating the monthly aggregates in a way no later valid run could correct.
- No "sensible fallback" window. An invalid setting carries no signal about intent; an admin who meant 365 should not silently lose 275 days because the code guessed 90. It refuses and logs.
- `rollupAndPurge()` takes `$purge`; the plugin passes `enable_purge`.
- Plugin validates the raw param with `is_numeric` before casting — the only layer that can still distinguish a blank field from a deliberate `0`.

## Provenance

This is not hypothetical. The same class of bug destroyed 74 rows in a developer database during this work.

bcordis:
Fixes #1571
Fixes #1608

## 1. `ltrim($host, 'www.')` corrupted w-initial domains

`ltrim`'s second argument is a **character mask**, not a prefix. Any host beginning with `w`, `.` or `3` lost leading characters:

| input | stored |
|---|---|
| `worship.example.org` | `orship.example.org` |
| `watch.church.tv` | `atch.church.tv` |
| `wordoflife.org` | `ordoflife.org` |

That corrupted `referrer_domain` and broke the internal-vs-external host comparison in `classifyReferrer()`. Replaced with an explicit `stripWwwPrefix()` at all three call sites.

## 2. Rollup was not atomic

The aggregate INSERT and the purge DELETE were independent statements. A task killed between them left the raw events in place with the aggregates already written, so the next run counted them again — permanently inflating monthly totals with no self-correction. Both now sit in one transaction.

**The savepoint flag on those calls is load-bearing.** `MysqliDriver::transactionStart()` calls `begin_transaction()` unconditionally when not asked for a savepoint, and **MySQL implicitly COMMITs any in-progress transaction on BEGIN** — so a bare call silently commits a caller's open transaction.

This is not theoretical. A bare call here committed the test suite's own isolating transaction and **permanently deleted 74 rows from a developer database** before the cause was identified. `tearDown()` now snapshots the table and fails loudly if a rollback ever fails to restore it, so a future isolation break is a red test rather than silent data loss.

## 3. `uq_aggregate` never matched

The unique key spans five NULLable columns, and **MySQL treats NULL as distinct from NULL in a unique index** — so `ON DUPLICATE KEY UPDATE` never fired and every run inserted duplicates instead of consolidating. Confirmed directly against MySQL with a probe table (3 rows where 1 was expected). Switched to the delete-then-reinsert approach `seedFromLegacy()` already uses for exactly this reason.

`rolled` now reports monthly groups written; `getAffectedRows()` counts 1 per insert but 2 per update and was never an accurate count of anything.

## 4. Purge safety (#1608)

`(int) ($params->retention_days ?? 90)` only defended against NULL:

| stored | days | cutoff | effect |
|---|---|---|---|
| absent | 90 | now −90d | correct |
| `""` blank field | 0 | **now** | deletes every event |
| `"abc"` | 0 | **now** | deletes every event |
| `"-5"` | −5 | **now +5d** | deletes every event |

⚠️ **`filter="int"` on the form field would NOT fix this** — it turns `""` into exactly the lethal `0`. Please don't "fix" it that way later.

An out-of-range window now aborts **both** halves, not just the delete: `$retentionDays` *is* the cutoff, so `0` would also have the rollup summarise today's events while the raw rows survived — inflating aggregates in a way no later valid run could correct.

It refuses and logs rather than substituting a default. An invalid setting carries no signal about intent, and an admin who meant 365 should not silently lose 275 days to a guess.

The floor lives **at the point of deletion**, so it covers restored backups (`Cwmbackup` rewrites `#__scheduler_tasks` wholesale), migrations and direct DB edits — not just the one caller.

`enable_purge` was read by the plugin and then **ignored**, because `rollupAndPurge()` deleted unconditionally. An admin who set **Purge: No** still lost their raw events. Now honoured, with the purge inside the rollup's transaction against the same cutoff — so no raw event can be deleted that this run did not aggregate first.

`isPurgeSafe()` is public and pure so the guard is provable **without a database**. Testing it through `rollupAndPurge()` would run the real DELETE if the guard were wrong.

## 5. GeoIP docblock corrected

`logEvent()` claimed a GeoIP classification that exists nowhere in the codebase — `country_code` is absent from the INSERT and is always NULL. Left unimplemented rather than wired up: it means a new dependency and a new privacy surface against a schema that documents never storing the raw IP, and no admin view renders a country breakdown.

## Testing

24 tests in `CwmanalyticsRollupTest`; full suite **1514 tests / 6026 assertions** green. Table row counts verified externally as identical before and after the run (33 events / 2471 monthly).

## Follow-up

#1609 — the component's own `analytics_retention_days` setting is dead; nothing reads it.


bcordis:
Added: **analytics history is now kept for all time by default.** The purge defaults to OFF in both the task form and the plugin fallback, making retention opt-in.

The evidence for this being the right default, rather than a preference: `CwmanalyticsModel` reads `#__bsms_analytics_events` in ~20 queries — every breakdown, trend, top-content and drill-down panel — but reads `#__bsms_analytics_monthly` in exactly **one**, `getLegacyKpiTotals()`, which renders as its own all-time widget rather than backing the main KPIs. The monthly aggregates do not stand in for the raw events, so a purge blanks most of the dashboard for that period instead of compacting it.

Both defaults changed because they are independent: the form default governs tasks configured in the Scheduler UI, while the plugin fallback governs tasks the setup wizard seeds with `params = '{}'` — which have been falling through to `?? true` and purging without anyone choosing it. Those installs stop purging on update. That is a behaviour change, but only ever toward retaining data.

`retention_days` is now hidden behind `showon="enable_purge:1"` so it no longer implies a limit that is not being applied.

**One caveat, filed as #1611 rather than fixed here:** `referrer_url`, `referrer_domain` and `session_hash` are marked *Consent-required* in the schema, and keeping those indefinitely sits awkwardly with GDPR storage limitation — which this component otherwise honours (DNT, opt-out cookie, raw IP never stored). The fix is to expire those three columns rather than delete whole rows, which keeps every dimension the dashboard reports on while the personal data still ages out. Better on both counts than the current purge, which discards both.

Suite now 1516 tests / 6029 assertions green. The two new default assertions were red/green verified by reverting each default independently.

Also: the earlier `Analyze (javascript-typescript)` failure was a GitHub infrastructure outage (`Bad Gateway` resolving `actions/checkout@v7`), not this branch — the same job passed on a parallel run.

PR #1616, opened 2026-08-06
fix: key and daily-rotate the analytics session hash

### LINKED ISSUE #1611: Analytics: expire the consent-required columns instead of deleting whole event rows
Follow-up to #1608, where the analytics purge was defaulted to OFF so history is kept for all time.

That is the right default — `CwmanalyticsModel` reads `#__bsms_analytics_events` in ~20 queries but `#__bsms_analytics_monthly` in exactly one (`getLegacyKpiTotals()`), so purging blanks most of the dashboard rather than compacting it.

## The one thing that argues against unlimited retention

Three columns in `#__bsms_analytics_events` are explicitly marked as personal data in the schema:

```sql
`referrer_url`    VARCHAR(2048) NULL COMMENT 'Consent-required',
`referrer_domain` VARCHAR(255)  NULL COMMENT 'Consent-required',
`session_hash`    VARCHAR(64)   NULL COMMENT 'SHA-256 of session ID; consent-required',
```

Keeping those indefinitely sits awkwardly with the GDPR storage-limitation principle, which the component otherwise takes seriously — DNT header support, an opt-out cookie, raw signals discarded at log time, and the raw IP never stored at all.

So "keep everything forever" is right for the *analytics*, and wrong for the *personal data* riding along in the same rows.

## Proposal: anonymise rather than delete

Add an expiry that `NULL`s the three consent-required columns on events older than N days, leaving the row intact.

This keeps **every dimension the dashboard actually reports on** — `study_id`, `series_id`, `media_id`, `location_id`, `event_type`, `referrer_type`, `device_type`, `browser`, `os`, `language`, `country_code`, `created` — while the personal data expires on schedule. `referrer_type` (the enum: direct/organic/social/email/internal/other) survives, so referrer reporting is unaffected; only the specific URL and domain go.

Net effect: better GDPR posture *and* better analytics than the current purge, which throws away both.

## Notes

- Should be a third toggle on the analytics task, defaulting **on** with a sensible window, since unlike the purge it is not lossy for reporting.
- `country_code` is always NULL today (no GeoIP anywhere in the codebase — see #1571), so it is not a concern yet, but would become one if GeoIP is ever added.
- The `MIN_RETENTION_DAYS` floor and `isPurgeSafe()` guard from #1608 should apply to this window too — an anonymisation window of 0 would wipe the columns for all time in one run.

bcordis:
Fixes #1611

## The problem

```php
$sessionHash = hash('sha256', $sessionId);   // unkeyed, and stable forever
```

Two consequences:

1. **Unkeyed** — anyone holding a candidate session ID could reproduce the digest and confirm a visitor's presence.
2. **Stable forever** — one visitor could be followed across the entire history of the table. **Measured on a real 82,565-row site: 188 visitors were linkable across days.**

## The fix

Keyed with Joomla's site secret, with the UTC day folded into a domain-separated subkey:

```php
$dailyKey = hash_hmac('sha256', self::SESSION_HASH_CONTEXT . ':' . $day, $secret);
return hash_hmac('sha256', $sessionId, $dailyKey);
```

| | Before | After |
|---|---|---|
| Same visitor, same day | same hash | **same hash** (Sessions KPI unaffected) |
| Same visitor, next day | same hash | **different hash** (unlinkable) |
| Reproducible without the secret | yes | no |

Mirrors the rotating-salt approach Plausible and Fathom use — the basis on which they describe themselves as storing no personal data.

## No new configuration

@bcordis asked whether session keys are Joomla-driven. They are, and that made this cheaper than the option I originally quoted (which said "new param for the secret" — not needed):

- **Session ID** is Joomla's: `$app->getSession()->getId()`. Proclaim creates no session, sets no cookie, controls no expiry.
- **Key** is Joomla's: `$app->get('secret')`, generated at install by `UserHelper::genRandomPassword(16)` and already used by `Version.php`.
- **Rotation** comes from folding the date into the derivation, not from rotating anything stored — so there is nothing to schedule, manage or back up.

**Accuracy cost is nil in practice**: Joomla's default session lifetime is **15 minutes**, so sessions effectively never span a day boundary for the rotation to truncate. (This also explains the earlier finding that 98.4% of sessions hold a single event — that is genuine single-page sermon visits, not a broken identifier.)

## Fail-safe

Where no site secret is available the hash is left **NULL** rather than falling back to an unkeyed digest. Losing a row from the Sessions count is preferable to writing a permanently linkable identifier.

## Migration — validated against the real export

`10.5.6-20260806.sql` re-keys existing rows as `SHA2(CONCAT(session_hash, DATE(created)), 256)`, giving history the same cross-day unlinkability. Run against the production export:

| Check | Result |
|---|---|
| Days where the Sessions figure moved | **0** |
| Hashes linkable across days | **188 → 0** |
| NULL hashes preserved | 1,156 (exact — `CONCAT()` propagates NULL) |
| Digests still well-formed 64-hex | 81,409 (all) |

## Honest scope

This is a **large reduction in linkability, not anonymisation.** The value remains pseudonymous within a single day and is still consent-required. Per #1611's measurements, only aggregation reaches anonymity — expiring columns still left 67.9% of rows uniquely identifying at day resolution.

## Testing

8 new tests in `CwmanalyticsSessionHashTest`, including a behavioural check that the migration's re-key preserves per-day `COUNT(DISTINCT)` — run against a `TEMPORARY` table so there is no path from the test to real rows. Red-checked by reverting the derivation to the bare hash: three assertions fail.

Full suite **1536 tests / 6058 assertions** on PHP 8.5.7. Dev analytics tables verified unchanged before and after.

⚠️ CI may fail on `Set up job` — GitHub Actions is [under an active incident](https://www.githubstatus.com/). Locally verified on 8.3.31 / 8.5.3 / 8.5.7.

### DIFF
--- a/admin/sql/install.mysql.utf8.sql
+++ b/admin/sql/install.mysql.utf8.sql
@@ -1193,7 +1193,7 @@ CREATE TABLE IF NOT EXISTS `#__bsms_analytics_events` (
     `os`              VARCHAR(50) NULL DEFAULT NULL,
     `language`        VARCHAR(10) NULL DEFAULT NULL,
     `is_guest`        TINYINT(1) NULL DEFAULT NULL COMMENT '0=logged in, 1=guest',
-    `session_hash`    VARCHAR(64) NULL DEFAULT NULL COMMENT 'SHA-256 of session ID; consent-required',
+    `session_hash`    VARCHAR(64) NULL DEFAULT NULL COMMENT 'Keyed hash of session ID, rotated daily; not linkable across days; consent-required',
     `created`         DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
     PRIMARY KEY (`id`),
     KEY `idx_study_created`    (`study_id`, `created`),
--- a/admin/sql/updates/mysql/10.5.6-20260806.sql
+++ b/admin/sql/updates/mysql/10.5.6-20260806.sql
@@ -24,3 +24,21 @@ ALTER TABLE `#__bsms_studytopics`
 -- them.
 DELETE FROM `#__bsms_studytopics`
 WHERE `study_id` NOT IN (SELECT `id` FROM `#__bsms_studies`);
+
+-- #1611: session_hash is now a keyed digest rotated daily, so a visitor cannot
+-- be followed across days. Update the column comment to say so.
+ALTER TABLE `#__bsms_analytics_events`
+    MODIFY `session_hash` VARCHAR(64) NULL DEFAULT NULL
+    COMMENT 'Keyed hash of session ID, rotated daily; not linkable across days; consent-required';
+
+-- Re-key the existing history. Rows written before this release hold a bare,
+-- unsalted SHA-256 that is stable for all time, so one visitor could be traced
+-- across the whole table. Folding each row's own date into the digest gives
+-- those rows the same cross-day unlinkability new rows get.
+--
+-- Rows from the same visitor on the same day still collide, so
+-- COUNT(DISTINCT session_hash) is unchanged and no historical Sessions figure
+-- moves. NULL stays NULL because CONCAT() propagates it.
+UPDATE `#__bsms_analytics_events`
+SET `session_hash` = SHA2(CONCAT(`session_hash`, DATE(`created`)), 256)
+WHERE `session_hash` IS NOT NULL;
--- a/admin/src/Helper/CwmanalyticsHelper.php
+++ b/admin/src/Helper/CwmanalyticsHelper.php
@@ -145,13 +145,28 @@ class CwmanalyticsHelper
             // Campus: resolved from study or media record
             $locationId = self::resolveLocationId($studyId, $mediaId);
 
-            // Session hash (personal data — consent-required)
+            // Session hash (personal data — consent-required).
+            //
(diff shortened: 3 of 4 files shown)


bcordis:
Renamed `deriveSessionHash` → **`hashSessionForDay`** on review feedback — "derive" described the mechanism rather than the purpose and read as opaque.

The docblock now leads with *why the column exists* (counting distinct visits without recording who made them) before getting to the keying and rotation. The previous opening assumed the reader already knew what the value was for.

No behavioural change; the derivation itself is untouched. Suite still 1536 tests / 6058 assertions.

bcordis:

bcordis:
Rebased onto `development` after #1620 merged. Both branches added `admin/sql/updates/mysql/10.5.6-20260806.sql`; the two sets of statements are now folded into that one file, per the one-file-per-unreleased-version convention.

Rehearsed the combined migration against copies of the real tables: 68 duplicate topic pairs to zero, `uq_study_topic` created, `session_hash` comment applied, no errors. `install.mysql.utf8.sql` merged cleanly — the two changes touch different tables.

Commits re-signed after the rebase (`development` requires signed commits, and a rebase drops signatures).

PR #1618, opened 2026-08-06
fix: stop overwriting admin-edited guided tours on every update

### LINKED ISSUE #1572: CwmguidedtourHelper: every Proclaim update unconditionally overwrites admin-edited tour/steps; TOCTOU race creates permanently orphaned duplicate tours

## 1. `updateTour()` unconditionally clobbers admin-edited tour/step state on every Proclaim install or update

`admin/src/Helper/CwmguidedtourHelper.php:676-748`. `registerguidedtours` is a step in both finish-step lists in `CwminstallModel.php` (full migration AND the minimal no-DB-changes list) — it runs on every Proclaim release, including no-op patch bumps. If an admin has edited the tour via Joomla core's Guided Tours UI (unpublished it, raised its access level, edited step wording, added a translation), the next update's `updateTour()` overwrites `title`/`description`/`url`/`published`/`access`/`language`/`note`/`autostart` from the hardcoded array, re-publishing at `access=1` regardless — and `DELETE`s every step row, reinserting the 12 hardcoded steps, destroying any admin edits or translated rows. Deterministic on every update, logged only as an unremarkable INFO message.

**Fix direction:** Don't treat the tour as fully owned/overwritten data on every update. Either version-gate re-registration (stored hash/version column) or only insert-if-missing, exposing a separate explicit "reset to default" action.

## 2. Missing UNIQUE constraint on `uid` lets a TOCTOU race create permanently orphaned duplicate tours

`CwmguidedtourHelper.php:479-501,530-543,794-830`. Core's `#__guidedtours` table defines `uid` with only a non-unique `KEY`, not UNIQUE. `registerGuidedTours()` does check-then-act (`tourExists()` then `insertTour()`/`updateTour()`) with no transaction/locking. Two concurrent invocations (overlapping admin requests during an update) can both see `tourExists() === false` and both insert, producing two rows with the identical `uid`, each with its own 12 duplicate steps. Damage is permanent: `getTourId()` has no `ORDER BY`/`LIMIT`, returning an arbitrary duplicate (used for autostart wiring and every future `updateTour()`); `removeAllTours()` also does a plain `loadResult()` per `uid`, so uninstall deletes only one duplicate — the other survives in core Joomla tables forever, even past a full Proclaim uninstall.

**Fix direction:** Change lookups to `loadColumn()` and handle the multi-row case (operate on the lowest/highest `id` deterministically; `removeAllTours()` should loop over all matches). Longer-term, add a UNIQUE index migration for `uid` or serialize registration with a lock/transaction.

bcordis:
Fixes #1572

## 1. Every update rewrote the tour

`registerGuidedTours()` rewrote any tour it found from the hardcoded definition. It is a finish step in **both** of `CwminstallModel`'s lists — the full migration *and* the minimal no-DB-changes one (lines 362 and 371) — so it ran on **every Proclaim release, including no-op patch bumps**.

Each run reset `title`, `description`, `url`, `published`, `access`, `language`, `note` and `autostart`, then `DELETE`d every step row and reinserted the twelve shipped ones.

An administrator who unpublished the tour, raised its access level, reworded a step, or added translated step rows lost all of it on the next update — silently, with an unremarkable INFO log line.

**An installed tour now belongs to the site.** Registration inserts when absent and otherwise leaves it alone. The destructive refresh remains available via `resetGuidedTours()`, so restoring shipped defaults is something a site *asks for* rather than something an update does to it.

## 2. Duplicate tours were permanent

Verified: core indexes `uid` with a plain `KEY idx_uid (uid(191))`, **not** UNIQUE — in `extensions.sql` and against a live database (`non_unique=1`). Registration is check-then-act with no lock, so two overlapping requests could both find the tour missing and both insert.

The surplus rows were permanent:

- `getTourId()` used `loadResult()` with no `ORDER BY` → returned an arbitrary duplicate, which then drove autostart wiring and every future update
- `removeAllTours()` deleted **one row per uid** → the rest survived in **core tables even after Proclaim was uninstalled**

Now: lookups go through `getTourIds()` ordered by `id`; `removeAllTours()` deletes every match; and `insertTour()` prunes surplus rows immediately after inserting — which also **heals databases that already have duplicates**.

### Deliberately no UNIQUE index

The issue suggested a UNIQUE migration on `uid` as a longer-term fix. I did not do that: `#__guidedtours` is a **core Joomla table**, and altering its schema from an extension puts us in conflict with core migrations and with any other extension registering tours. Converging after the fact is the safer half of that trade.

## Testing

8 tests in `CwmguidedtourHelperTest`, run against the real core tables inside a savepoint-aware transaction, with a `tearDown()` row-count assertion — these write to *core* tables, so a broken isolation boundary would damage more than Proclaim's own data. Verified 16 tours / 132 steps unchanged before and after.

Tests scope the helper to a synthetic tour via reflection. My first version asserted against the shipped uid and failed, because a real site already carries that row — the tests were really asserting against whatever the site happened to have.

**Red-checked per fix**, and two of my first attempts were invalid, so I redid them:

| Fix | Reverted → |
|---|---|
| Preserve on update | 2 tests fail |
| `removeAllTours` multi-delete | 1 test fails |
| Deterministic ordering | **0 tests fail** |

That last one is honest rather than fixed: without `ORDER BY`, MySQL returns rows in primary-key order anyway, so a runtime assertion passes either way. It is pinned with a source-level assertion instead of claiming coverage that does not exist.

Full suite **1544 tests / 6070 assertions**.

⚠️ CI may fail at `Set up job` — GitHub Actions incident is ongoing. Verified locally on PHP 8.5.7.

### DIFF
--- a/admin/src/Helper/CwmguidedtourHelper.php
+++ b/admin/src/Helper/CwmguidedtourHelper.php
@@ -476,7 +476,7 @@ class CwmguidedtourHelper
      *
      * @since 10.1.0
      */
-    public function registerGuidedTours(): int
+    public function registerGuidedTours(bool $force = false): int
     {
         if (!$this->supportsGuidedTours()) {
             Log::add('Guided tours not supported on this Joomla version', Log::INFO, 'com_proclaim');
@@ -486,18 +486,53 @@ class CwmguidedtourHelper
         $count = 0;
 
         foreach ($this->tours as $key => $tour) {
-            if ($this->tourExists($tour['uid'])) {
-                if ($this->updateTour($key, $tour)) {
+            if (!$this->tourExists($tour['uid'])) {
+                if ($this->insertTour($key, $tour)) {
                     $count++;
-                    Log::add('Updated guided tour: ' . $key, Log::INFO, 'com_proclaim');
+                    Log::add('Registered guided tour: ' . $key, Log::INFO, 'com_proclaim');
                 }
-            } elseif ($this->insertTour($key, $tour)) {
+
+                continue;
+            }
+
+            // The tour is already present. Once installed it belongs to the
+            // site: an administrator may have unpublished it, raised its access
+            // level, reworded steps or added translated rows through Joomla's
+            // Guided Tours UI. Rewriting it here would discard all of that, and
+            // this method runs on every Proclaim release including no-op patch
+            // bumps, so the loss would be silent and repeated.
+            //
+            // Refreshing from the shipped definition therefore has to be asked
+            // for explicitly.
+            if (!$force) {
+                Log::add('Preserved existing guided tour (admin-owned): ' . $key, Log::INFO, 'com_proclaim');
+
+                continue;
+            }
+
+            if ($this->updateTour($key, $tour)) {
                 $count++;
-                Log::add('Registered guided tour: ' . $key, Log::INFO, 'com_proclaim');
+                Log::add('Reset guided tour to shipped default: ' . $key, Log::INFO, 'com_proclaim');
             }
         }
 
         return $count;
+    }
+
+    /**
+     * Discard local changes and rewrite every tour from its shipped definition.
+     *
+     * Destructive: replaces tour metadata and deletes and reinserts all steps,
+     * losing admin edits and translated step rows. Exposed so a reset can be
+     * offered deliberately, rather than happening as a side effect of updating.
+     *
+     * @return  int  Number of tours rewritten
+     *
+     * @since   __DEPLOY_VERSION__
+     */
+    public function resetGuidedTours(): int
+    {
+        return $this->registerGuidedTours(true);
     }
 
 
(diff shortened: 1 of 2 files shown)


bcordis:
**Corrected the branch — this PR was carrying three commits that belong to #1616.**

I branched `1572-guidedtour-preserve-and-dedupe` from `1611-analytics-session-hash-rotation` instead of `development`, so this PR was showing four commits and touching `CwmanalyticsHelper.php`, `install.mysql.utf8.sql` and the 10.5.6 migration — none of which have anything to do with guided tours.

Rebased onto `development` and force-pushed. It is now a single commit touching exactly two files:

```
admin/src/Helper/CwmguidedtourHelper.php
tests/integration/Admin/Helper/CwmguidedtourHelperTest.php
```

The rebase dropped the commit signature, so it has been re-signed — `development` requires signed commits, and an unsigned one would have made this unmergeable. Verified signed, and all 8 guided-tour tests still pass.

No change to the actual fix; only which commits this PR proposes to merge.

bcordis:

PR #1619, opened 2026-08-06
fix: make scripture reference saving atomic

### LINKED ISSUE #1573: CwmscriptureHelper: saveScriptures() delete+insert not transactional — mid-loop failure permanently loses references and desyncs legacy flat columns

## `saveScriptures()` deletes then inserts with no transaction — a mid-loop failure permanently loses existing references and desyncs the legacy flat columns

`admin/src/Helper/CwmscriptureHelper.php:157-193`. `saveScriptures($studyId, $scriptures)` issues a `DELETE FROM #__bsms_study_scriptures WHERE study_id = X`, then loops inserting each new reference one row at a time — no `transactionStart()`/`transactionCommit()` wrapping the two (`#__bsms_study_scriptures` is InnoDB, so a transaction is viable). If the insert loop throws partway through (DB error, connection blip, deadlock on any insert after the first), the exception propagates before all references are written, and because callers (`CwmmessageModel.php:577-578`, `CwmcsvimportHelper.php:247-248,345-346`) call `saveScriptures()` immediately followed by `syncLegacyColumns()`, that follow-up call never runs either. Net result: old references gone, only some new ones landed, and the legacy flat columns on `#__bsms_studies` (`booknumber`, `chapter_begin`, etc.) still describe the *old* reference set — a permanent divergence nothing will reconcile until the next successful save.

**Fix direction:** Wrap the delete + insert loop (and ideally the subsequent `syncLegacyColumns()` call at the caller sites) in `$db->transactionStart()`/`transactionCommit()` with `transactionRollback()` on exception.

bcordis:
Fixes #1573

## The reference set could be left half-written

`saveScriptures()` deleted a study's reference rows and then inserted the new set one at a time, with nothing wrapping the two:

```php
$db->execute();                    // DELETE all rows for the study
foreach ($scriptures as $i => $ref) {
    $db->execute();                // INSERT, one at a time
}
```

An insert failing partway through — a bad value, a deadlock, a dropped connection — left the study with its **previous references already deleted and only some of the new ones written**. Nothing reconciled that afterwards.

Confirmed `#__bsms_study_scriptures` is InnoDB, so a transaction is viable.

The exception still reaches the caller, so upstream error handling is unchanged. What changes is that the rows are back the way they were when it does.

## And a second half: the flat columns diverged

`#__bsms_studies` carries a flattened copy of the first two references. All three call sites called the two operations separately:

```php
CwmscriptureHelper::saveScriptures($studyId, $scriptures);
CwmscriptureHelper::syncLegacyColumns($studyId, $scriptures);
```

When the first threw, the second never ran — so the flat columns went on describing the **previous** reference set while the junction table held something else. Permanent, until some later save happened to succeed.

`saveScripturesAndSync()` now writes both in one transaction, and all three call sites (`CwmmessageModel`, and both CSV import paths) use it.

## Savepoint-aware, deliberately

`MysqliDriver::transactionStart()` calls `begin_transaction()` unconditionally without the flag, and **MySQL implicitly commits any transaction already open** when it does — so a bare call here would commit a caller's transaction out from under it. No caller opens one today, but saving a study is exactly the kind of operation that grows one later. (This is not hypothetical; the same mistake destroyed 74 rows in a developer database earlier in this sweep.)

## Testing

5 tests that provoke a **real** mid-loop failure — an invalid value on the second reference — rather than mocking the driver or asserting on source text.

Both failure paths red-checked by reverting the transactions:

| Test | Reverted → |
|---|---|
| Original references restored | `Failed asserting that 1 is identical to 3` |
| No divergence after failure | `Failed asserting that two arrays are identical` |

⚠️ The second of those was initially a **weak assertion I had to strengthen**: `assertSame(1, countReferences())` passed in both states, because post-fix the single row is the restored reference and pre-fix it is an orphan from the failed write — same count, opposite meaning. It now asserts the surviving `booknumber`.

Fixture creates its own study, runs inside a savepoint-aware transaction, and `tearDown()` asserts the junction table row count is restored. Verified 827 studies / 441 references unchanged before and after.

Full suite **1533 tests / 6056 assertions**.

⚠️ CI may fail at `Set up job` — the GitHub Actions incident is ongoing. Verified locally on PHP 8.5.7.

### DIFF
--- a/admin/src/Helper/CwmcsvimportHelper.php
+++ b/admin/src/Helper/CwmcsvimportHelper.php
@@ -244,8 +244,7 @@ class CwmcsvimportHelper
             $scriptures = self::parseScriptures($rowData['scripture']);
 
             if (!empty($scriptures)) {
-                CwmscriptureHelper::saveScriptures($studyId, $scriptures);
-                CwmscriptureHelper::syncLegacyColumns($studyId, $scriptures);
+                CwmscriptureHelper::saveScripturesAndSync($studyId, $scriptures);
             }
         }
 
@@ -342,8 +341,7 @@ class CwmcsvimportHelper
             $scriptures = self::parseScriptures($rowData['scripture']);
 
             if (!empty($scriptures)) {
-                CwmscriptureHelper::saveScriptures($studyId, $scriptures);
-                CwmscriptureHelper::syncLegacyColumns($studyId, $scriptures);
+                CwmscriptureHelper::saveScripturesAndSync($studyId, $scriptures);
             }
         }
 
--- a/admin/src/Helper/CwmscriptureHelper.php
+++ b/admin/src/Helper/CwmscriptureHelper.php
@@ -160,6 +160,88 @@ class CwmscriptureHelper extends ScriptureHelper
 
         $db = Factory::getContainer()->get(DatabaseInterface::class);
 
+        // The delete and the inserts are one unit. Without that, an insert
+        // failing partway through leaves the study with its old references
+        // gone and only some of the new ones written, which nothing later
+        // reconciles.
+        //
+        // Savepoint-aware so this can nest inside a caller's transaction:
+        // MysqliDriver::transactionStart() calls begin_transaction()
+        // unconditionally otherwise, and MySQL implicitly commits any
+        // transaction already open when it does.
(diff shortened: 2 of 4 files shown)


bcordis:

Issue #1566, opened 2026-08-05
CwmdbHelper: performDB()/resetdb() assume execute() returns false on failure (it throws, breaking migration-rollback recovery); getObjects() unanchored substring match weakens export allowlist


bcordis:
Found during a Helper-folder-wide bug-hunting review, tier 3.

## 1. `performDB()`/`resetdb()` treat `$db->execute()` as if it returns `false` on failure — it never does

`admin/src/Helper/CwmdbHelper.php:226-247,471-535`. Both methods gate error handling on `if (!$db->execute())`. Joomla's `DatabaseDriver::execute()` never returns `false` — on success it returns `true`; on failure it throws `ExecutionFailureException`. The `false`-branch in both methods is dead code, and every real SQL failure surfaces as an uncaught exception instead of the documented "returns false" contract.

Concrete path: `CwminstallModel::realRun()` calls `CwmdbHelper::resetdb()` as its migration-rollback recovery step when an update step fails. If any statement inside `resetdb()`'s uninstall/install SQL fails (plausible during a rollback-of-a-failed-migration, e.g. a table left half-created erroring on "already exists"), `resetdb()` throws instead of returning `false`, propagating straight out of `realRun()`/`run()` (neither has a surrounding try/catch) — `resetStack()` never runs and `saveStack()` is skipped, leaving a stale/dirty migration stack and a fatal error instead of the intended graceful "not migrated" warning. Second path: `CwmadminController::copyTables()` calls `performDB()` three times per table with no try/catch, expecting `false` to trigger a clean abort; a failing `CREATE TABLE ... LIKE` instead throws uncaught mid-loop.

**Fix direction:** Wrap `$db->execute()` in `try { ... } catch (\Exception $e) { enqueue/log; return false; }` inside `performDB()` and `resetdb()`, matching the return-bool contract the docblocks already claim.

## 2. `getObjects()` uses unanchored substring matching instead of a prefix check, weakening a security allowlist

[...]

PR #1620, opened 2026-08-06
fix: make topic associations atomic, unique and cascade-deleted

### LINKED ISSUE #1574: CwmstudytopicHelper: non-atomic save loses topic associations on failure, TOCTOU duplicate junction rows (corroborated by existing cleanup tooling), missing delete-cascade orphans rows

## 1. No transaction around delete+insert — partial `store()` failure permanently loses topic associations

`admin/src/Helper/CwmstudytopicHelper.php:70-106`. `saveTopics()` deletes all existing topic rows for a study, then loops inserting the new set. If a mid-loop `store()` call fails (DB deadlock, connection drop, validation failure), the method returns `false` with no rollback — the study is left with neither the original nor the intended topic set. The only caller, `CwmmessageModel::saveTopics()` (a `void`-returning private wrapper), discards the boolean entirely, so nothing surfaces beyond a queued admin notice — the overall message save still reports success.

**Fix direction:** Wrap the delete + insert loop in a transaction so a mid-loop failure restores the prior association set instead of a partial mix.

## 2. No unique constraint + no locking — concurrent saves duplicate junction rows

`CwmstudytopicHelper.php:70-106`. `#__bsms_studytopics` has only a non-unique `KEY idx_study_topic (study_id, topic_id)`, no UNIQUE constraint. Two concurrent saves for the same study (double-click Save, two browser tabs) both race the DELETE and both INSERT, producing duplicate junction rows with no constraint to prevent it. **Independently corroborated**: `CwmdbHelper::cleanStudyTopics()` already exists as maintenance tooling specifically to delete duplicate `#__bsms_studytopics` rows — evidence this race has actually occurred in production.

**Fix direction:** Wrap the delete+insert in a transaction with row locking, or add a genuine `UNIQUE (study_id, topic_id)` index and use upsert semantics instead of unconditional `store()`.

## 3. Missing delete-cascade — deleting a study leaves permanently orphaned junction rows

The file's own docblock contrasts itself with `CwmstudyteacherHelper`, which has a `deleteTeachers(int $studyId)` method wired into `CwmmessageTable::delete()`. `CwmstudytopicHelper` has no equivalent `deleteTopics()`, and `CwmmessageTable::delete()` cleans up scripture refs and teacher associations but never touches `#__bsms_studytopics`. Every study/message deletion leaves permanent orphan rows, joined against by `CwmschemaorgHelper.php:973`, `Cwmtranslated.php:141-145`, `TopicsListField.php:69`, `TopicsFormField.php:142` — corrupting topic usage counts and filter dropdowns with phantom associations over time.

**Fix direction:** Add `CwmstudytopicHelper::deleteTopics(int $studyId): void` mirroring `CwmstudyteacherHelper::deleteTeachers()`, called from `CwmmessageTable::delete()` alongside the existing cleanup.

bcordis:
Fixes #1574

## 1. Non-atomic save

`saveTopics()` deleted a study's topic rows then reinserted the new set, unwrapped. A failure partway through left the study with **neither** the original associations nor the intended ones — and because the only caller discards the boolean, **the message save still reported success**.

Now one savepoint-aware transaction. A failure logs and tells the administrator the topics were left unchanged, rather than passing silently.

## 2. No unique constraint

`#__bsms_studytopics` carried only a non-unique `KEY (study_id, topic_id)`, so two concurrent saves could both delete and both insert.

**This is not theoretical.** `CwmdbHelper::cleanStudyTopics()` already exists solely to remove duplicates after the fact, and the development database carried **68 duplicated pairs across 392 rows**.

The pair is now `UNIQUE`, in the install schema and via migration.

⚠️ **The migration de-duplicates before adding the index** — `ADD UNIQUE` fails outright on any site that already has duplicates, which is most of them. Rehearsed against a copy of the real table:

| | |
|---|---|
| Rows | 392 → **117** |
| Duplicate pairs | 68 → **0** |
| Unique index | created (`non_unique=0`) |

## 3. No delete cascade

`CwmmessageTable::delete()` cleaned up scripture references and teacher associations but **never topics**. Every study deletion left rows behind, still joined against by topic counts, the topic filter dropdowns (`TopicsListField`, `TopicsFormField`), `Cwmtranslated` and `CwmschemaorgHelper` — reporting associations to content that no longer exists.

`deleteTopics()` mirrors the existing `deleteTeachers()` and is called alongside it. The migration clears orphans already present.

## Two extra bugs found while working

- `if (!$db->execute())` was an **unreachable branch** — `DatabaseDriver::execute()` returns true or throws, never false — so the delete's error path had never run. (Same class as #1566.)
- `createTable()` can return **false** when the factory cannot resolve the class, at which point `$tagRow->study_id = …` is a **fatal**, not a handled failure. Now checked. This surfaced because the integration harness resolves `com_proclaim` through a LegacyFactory where `createTable()` does return false.

## Testing

6 tests; both behavioural fixes red-checked:

| Fix | Reverted → |
|---|---|
| Transaction | `testFailedSaveRollsBack` fails |
| Delete cascade | `testDeleteCascadeIsWiredIn` fails |

⚠️ **My first transaction red-check was invalid** — the edit broke syntax, so the 3 "failures" were parse errors, not the revert. Redone as a clean no-op revert, which fails exactly one test.

**Coverage gap stated plainly:** the *success* path of `saveTopics()` is not integration-tested, because this harness's `createTable()` returns false so no row can be written. I dropped those tests rather than leave them skipped — a skipped test is one that silently never runs anywhere, including CI. Fixtures seed rows directly instead.

Full suite **1534 tests / 6061 assertions**; 827 studies / 392 topic rows verified unchanged.

## Migration filename

`10.5.6-20260806.sql` — the same filename PR #1616 adds. If that merges first, this branch needs a rebase merging both statements into one file, per the one-file-per-version convention.


bcordis:

PR #1621, opened 2026-08-06
fix: make teacher associations atomic

### LINKED ISSUE #1575: CwmstudyteacherHelper: saveTeachers() delete+insert not transactional — mid-loop DB failure leaves study saved but teacher list truncated with a fatal error shown
Found during a Helper-folder-wide bug-hunting review, tier 3.

## `saveTeachers()` has no transaction — mid-loop failure leaves the study already committed but its teacher list truncated

`admin/src/Helper/CwmstudyteacherHelper.php:171-210`. `CwmmessageModel::save()` calls `parent::save($data)` first (commits the study row), then `saveTeachers()`. Inside: `DELETE ... WHERE study_id = X` runs and succeeds, then the loop issues one `INSERT` per teacher. If a mid-loop insert hits a transient DB error (lock-wait timeout, deadlock — plausible with a concurrent bulk CSV import touching the same junction table, or a dropped connection), `$db->execute()` throws an uncaught `ExecutionFailureException`, the loop aborts, and the exception propagates out of `saveTeachers()` and `CwmmessageModel::save()`. The study row was already saved (title/body/etc. persisted), but the admin sees a fatal 500 as if the save failed, and the junction table now holds only a partial teacher list — nothing rolled back, no transaction anywhere in the chain.

**Fix direction:** Wrap the delete and the insert loop in `$db->transactionStart()`/`transactionCommit()`, with `transactionRollback()` + rethrow/log on failure.

## Related, separate bug (noted for triage, not part of this file)
`CwmteacherModel::canDelete()` (`admin/src/Model/CwmteacherModel.php:505-538`) queries for studies referencing a teacher but falls through to an authorization check instead of actually blocking deletion when studies exist — the "cannot delete" message is cosmetic. `CwmteacherTable::delete()` has no cascade cleanup of `#__bsms_study_teachers` at all (unlike `CwmmessageTable::delete()`, which correctly cleans up on the study side), leaving orphaned rows pointing at deleted teachers.

bcordis:
Fixes #1575

## The reported half

`saveTeachers()` deleted a study's teacher rows then inserted the new set one at a time, unwrapped. `CwmmessageModel::save()` **commits the study row first**, so a transient failure partway through left the study saved with its teacher list truncated — while the administrator saw a fatal as though nothing had been written.

Now one unit. The exception still reaches the caller, so upstream error handling is unchanged.

## A second half the issue did not mention

Three of the four call sites follow `saveTeachers()` with `syncLegacyColumn()`, which copies the primary teacher into `#__bsms_studies.teacher_id`:

```php
CwmstudyteacherHelper::saveTeachers($studyId, $teacherEntries);
CwmstudyteacherHelper::syncLegacyColumn($studyId, $teacherEntries);
```

When the first threw, the second never ran — leaving that column naming a teacher the junction table no longer held. `saveTeachersAndSync()` writes both in one transaction.

**This is the same divergence fixed for scripture references in #1573.** The two helpers were built to the same shape and carried the same defect, so it is worth assuming the pattern repeats wherever a junction table shadows a flat column.

The fourth call site — batch teacher assignment — sets `teacher_id` through the table before calling `saveTeachers()`, so it is already consistent and is left alone.

## Not a problem here, unlike #1574

`#__bsms_study_teachers` already carries a **UNIQUE** key on `(study_id, teacher_id)` — verified `non_unique=0` on a live database, with **zero** duplicate pairs. So this table has none of the duplication `#__bsms_studytopics` accumulated. No migration needed.

## Testing

5 tests provoking a real mid-loop failure via an out-of-range teacher id. Both failure paths red-checked by disabling the transactions:

| Test | Reverted → |
|---|---|
| Original list restored | fails |
| No legacy-column divergence | fails |

The divergence test asserts **which** teacher survived, not just how many — a count of one is satisfied both by the restored row and by an orphan from the failed write, which are opposite outcomes. (Same trap I had to fix in #1619.)

Full suite **1533 tests / 6055 assertions**; 827 studies / 826 junction rows verified unchanged.

## Also noted for triage

The issue flags a related bug in different files, which I have **not** fixed here and will file separately: `CwmteacherModel::canDelete()` enqueues a "cannot delete" message when studies reference the teacher, then **falls through to `return authorise('core.delete')`** — so the block is cosmetic and deletion proceeds. `CwmteacherTable::delete()` has no cascade cleanup of `#__bsms_study_teachers`, unlike `CwmmessageTable::delete()` on the study side.

### DIFF
--- a/admin/src/Helper/CwmcsvimportHelper.php
+++ b/admin/src/Helper/CwmcsvimportHelper.php
@@ -235,8 +235,7 @@ class CwmcsvimportHelper
                 static fn (int $id) => ['teacher_id' => $id],
                 $teacherIds
             );
-            CwmstudyteacherHelper::saveTeachers($studyId, $teacherEntries);
-            CwmstudyteacherHelper::syncLegacyColumn($studyId, $teacherEntries);
+            CwmstudyteacherHelper::saveTeachersAndSync($studyId, $teacherEntries);
         }
 
         // Save scriptures
@@ -333,8 +332,7 @@ class CwmcsvimportHelper
                 static fn (int $id) => ['teacher_id' => $id],
                 $teacherIds
             );
-            CwmstudyteacherHelper::saveTeachers($studyId, $teacherEntries);
-            CwmstudyteacherHelper::syncLegacyColumn($studyId, $teacherEntries);
+            CwmstudyteacherHelper::saveTeachersAndSync($studyId, $teacherEntries);
         }
 
         // Update scriptures if provided
--- a/admin/src/Helper/CwmstudyteacherHelper.php
+++ b/admin/src/Helper/CwmstudyteacherHelper.php
@@ -174,7 +174,84 @@ class CwmstudyteacherHelper
 
         $db = Factory::getContainer()->get(DatabaseInterface::class);
 
-        // Delete existing
+        // The delete and the inserts are one unit. CwmmessageModel::save()
+        // commits the study row before calling this, so a failure partway
+        // through the loop left the study saved but its teacher list
+        // truncated, with the administrator seeing a fatal as though nothing
+        // had been written.
+        //
+        // Savepoint-aware: MysqliDriver::transactionStart() calls
+        // begin_transaction() unconditionally otherwise, and MySQL implicitly
+        // commits any transaction already open when it does.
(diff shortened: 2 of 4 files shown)


bcordis:

PR #1624, opened 2026-08-06
fix: Cwmhelper fatal on null params, missed streaming skip, CLI warning

### LINKED ISSUE #1576: Cwmhelper: mediaBuildUrl() TypeError from a live null-passing caller (legacy player crash), getRemoteFileSize() streaming skip runs before URL normalization, unguarded HTTP_HOST breaks CLI scheduler

## 1. `mediaBuildUrl()` requires a non-nullable `Registry`, but a live caller passes `null`

`admin/src/Helper/Cwmhelper.php:199`. `Cwmmedia::getAVmediacode()` (`site/src/Helper/Cwmmedia.php:940,947`) is reached whenever a media record uses the legacy "All Videos Reloaded" player (player type 2/3, subtype 2 — inline playback). It calls `Cwmhelper::mediaBuildUrl($media->spath, $media->filename, null)`, passing literal `null` for the third `Registry $params` argument (no default, not nullable). Object/class type declarations are never weak-coerced — this is a hard `TypeError: Argument #3 ($params) must be of type Registry, null given` (verified). Any site with a message configured for this legacy inline player gets a fatal error instead of the video rendering.

**Fix direction:** Either make the parameter `?Registry $params = null` and guard the one dereference point (only reached when `$setProtocol` is true), or fix the caller to pass an actual `Registry` instance.

## 2. Streaming-host skip runs before URL normalization, so schemeless URLs miss the skip list

`Cwmhelper.php:82-88`. `getRemoteFileSize()` skips HTTP HEAD requests for streaming platforms (documented as "meaningless for embedded players"), computing `$host = parse_url($url, PHP_URL_HOST)` on the raw, un-normalized URL — before the function's own later scheme-normalization (lines 90-101). `parse_url()` only extracts a host when the input has a scheme or leading `//`; for a schemeless value like `"youtube.com/watch?v=abc123"` it returns `NULL` (verified). So the skip check compares against an empty string and never fires, and the function falls through to a real outbound HTTP HEAD request against YouTube — exactly the plausible, documented-as-expected data shape ("Removes a bad url problem in some DB's") the normalization block exists for.

**Fix direction:** Apply the same scheme-normalization the function already does later, *before* computing `$host` for the streaming-platform check.

## 3. Unconditional `$_SERVER['HTTP_HOST']` access breaks in the CLI-triggered scheduled task

`Cwmhelper.php:215`. `mediaBuildUrl()` unconditionally reads `$_SERVER['HTTP_HOST']` with no `isset()`/`??` guard. Called repeatedly from `Cwmpodcast::makePodcasts()`, invoked from the scheduled task plugin's `podcast()` routine. When that task runs via Joomla's CLI scheduler (`php cli/joomla.php scheduler:run`, the documented trigger for `com_scheduler`), there is no HTTP request and `$_SERVER['HTTP_HOST']` is undefined — a PHP `E_WARNING: Undefined array key` for every media file processed during feed regeneration. The block that consumes this value has no observable effect on the return value (the `$local` reassignment is discarded), so it's dead weight that also emits a warning.

**Fix direction:** Guard with `$host = $_SERVER['HTTP_HOST'] ?? '';`, or drop the block entirely since it has no downstream effect.

bcordis:
Fixes #1576

All three findings verified empirically before changing anything.

## 1. Fatal on a live call path

`mediaBuildUrl()` declared `Registry $params` — non-nullable, no default. `Cwmmedia::getAVmediacode()` passes literal `null` whenever a record uses the legacy inline player:

```php
Cwmhelper::mediaBuildUrl($media->spath, $media->filename, null)
```

**Object type declarations are never weak-coerced**, confirmed:

```
f(): Argument #1 ($p) must be of type ArrayObject, null given
```

So that is a fatal page instead of a rendered video for any site still using that player. Now `?Registry $params = null`, with the single dereference — reached only when `$setProtocol` is true — falling back to the same default it would have read.

## 2. The streaming skip never fired for schemeless URLs

`getRemoteFileSize()` skips streaming platforms because a file size is meaningless for an embedded player. But it computed the host from the **raw** URL, above the block that normalises a missing scheme:

| stored value | `parse_url(…, PHP_URL_HOST)` |
|---|---|
| `https://youtube.com/watch?v=abc` | `youtube.com` |
| `//youtube.com/x` | `youtube.com` |
| `youtube.com/watch?v=abc` | **NULL** |

So for a schemeless value the skip compared against `""`, never fired, and the function **made a real HEAD request to the streaming platform** — precisely the value shape the normalisation below it exists to repair ("Removes a bad url problem in some DB's"). Host is now read after normalisation.

## 3. Warning per media file under the CLI scheduler

`$_SERVER['HTTP_HOST']` was read unguarded. The podcast feed rebuild runs from Joomla's CLI scheduler, where no request exists — `E_WARNING: Undefined array key "HTTP_HOST"` for every media file processed.

The value fed a comparison whose only effect was reassigning `$local`, **which is never read again**. So the block is removed rather than guarded. `$local` stays in the signature because callers pass `$podcast` positionally after it; the docblock now says so explicitly rather than leaving a mystery parameter.

## Bonus

`if (empty($path)) { return false; }` in a method declared `: string`. In weak mode that was already silently becoming `""` — now the source says what actually happens.

## Testing

11 tests, all three findings red-checked:

| Fix | Reverted → |
|---|---|
| Nullable params | 4 tests fail |
| Host after normalisation | 3 tests fail — **exactly the schemeless cases** |
| HTTP_HOST guard | 1 test fails |

⚠️ **The streaming test needed strengthening.** My first version asserted only `=== 0`, which passes either way — a *failed* HEAD request also returns 0. It now asserts the private `$fileSizeCache` is untouched, which is written only once a lookup has been attempted. That makes an empty cache proof no request went out, and it means the passing test performs no network I/O. The reverted run genuinely hit the network; the fixed one does not.

Full suite **1539 tests / 6061 assertions** on PHP 8.5.7.

⚠️ CI may fail at `Set up job` — GitHub Actions incident ongoing.

### DIFF
--- a/admin/src/Helper/Cwmhelper.php
+++ b/admin/src/Helper/Cwmhelper.php
@@ -78,15 +78,6 @@ class Cwmhelper
             return 0;
         }
 
-        // Skip streaming platforms — file size is meaningless for embedded players
-        $host = (string) parse_url($url, PHP_URL_HOST);
-
-        foreach (self::$streamingHosts as $streamHost) {
-            if (str_contains($host, $streamHost)) {
-                return 0;
-            }
-        }
-
         // Removes a bad url problem in some DB's
         if (substr_count($url, '/http')) {
             $url = ltrim($url, '/');
@@ -97,6 +88,19 @@ class Cwmhelper
                 $url = 'https:' . $url;
             } else {
                 $url = 'https://' . $url;
+            }
+        }
+
+        // Skip streaming platforms — file size is meaningless for embedded
+        // players. This has to come after the normalisation above:
+        // parse_url() only reports a host when the value carries a scheme, so
+        // a stored "youtube.com/watch?v=..." yields no host and the skip never
+        // fires, sending a real HEAD request to a streaming platform instead.
+        $host = (string) parse_url($url, PHP_URL_HOST);
+
+        foreach (self::$streamingHosts as $streamHost) {
+            if (str_contains($host, $streamHost)) {
+                return 0;
             }
         }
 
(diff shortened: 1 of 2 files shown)


bcordis:

PR #1625, opened 2026-08-06
fix: stop one corrupted thumbnail crashing the public podcast list

### LINKED ISSUE #1577: Cwmimagelib: uncaught TypeError crashes public podcast page on a corrupted image (TypeError not caught by Exception handler), no dimension cap before decode (memory exhaustion)

## 1. Uncaught `TypeError` crashes the whole page when GD fails to decode a "valid" image

`admin/src/Helper/Cwmimagelib.php:76-84,138,154`. Empirically reproduced: a PNG whose header is intact but whose data stream is truncated passes `getimagesize()` cleanly (correct mime/width/height) but fails to decode — `imagecreatefrompng()` returns `false`. In `resizeImage()`, that `false` is passed to `imagecopyresampled()` (line 154), which type-hints `GdImage` in PHP 8 and throws a `TypeError` for a `bool` argument (also reproduced empirically). `TypeError` extends `\Error`, not `\Exception`, so `getSeriesPodcast()`'s `catch (\Exception $e)` never catches it. `getSeriesPodcast()` is called directly from `site/tmpl/cwmseriespodcastlist/default.php` for every item in the loop — one corrupted/truncated series-thumbnail image (interrupted upload, disk issue, bad backup restore) fatally crashes the entire public page for every visitor, instead of falling back to the original image as intended.

**Fix direction:** Validate the return of `image_create_func()` immediately and throw a catchable `\RuntimeException` (or widen the catch to `\Throwable`) before the value reaches `imagecopyresampled()`.

## 2. No dimension cap before full in-memory decode — decompression-bomb / memory-exhaustion risk

`Cwmimagelib.php:138-148`. `resizeImage()` decodes unconditionally regardless of pixel dimensions, which are already known from `getimagesize()` but never checked. PNG/GIF compress extremely well, so a small file can declare huge pixel dimensions (e.g. 20000×20000 ≈ 1.6GB decoded as RGBA). Decoding exhausts `memory_limit`, producing a raw PHP fatal that isn't a `Throwable` at all and can't be caught by any try/catch.

**Fix direction:** After `getimagesize()`, compare dimensions against a sane maximum (e.g. 4000×4000) and skip/fall back to the original image before calling `$image_create_func`.

bcordis:
Fixes #1577

## 1. A failed decode took down the whole page

Reproduced end to end:

| step | result |
|---|---|
| `getimagesize()` on a truncated PNG | **passes** — `200x150 image/png` |
| `imagecreatefrompng()` on it | **`false`** |
| `imagecopyresampled($tmp, false, …)` | **`TypeError`** |
| caught by `catch (\Exception)`? | **no** — TypeError extends `Error` |

`getSeriesPodcast()` is called per item from `site/tmpl/cwmseriespodcastlist/default.php`, so **one corrupted thumbnail crashed the entire public page for every visitor** — instead of falling back to the original image, which is exactly what that catch existed to do.

The decode result is now checked and a `RuntimeException` raised, which the caller already handles. The catch is widened to `Throwable` as a backstop.

## 2. No dimension cap before decode

Decoding happened before anything looked at the dimensions — even though `getimagesize()` had already reported them. PNG and GIF compress well enough that a small file can declare enormous dimensions, and **it is the decode that allocates**: 20000×20000 is roughly 1.6GB as RGBA.

Exceeding `memory_limit` is a **raw fatal, not a Throwable**, so no catch anywhere could have recovered. Sources beyond **8000px** on either edge are now refused before decoding; output is a 300×169 thumbnail, so nothing legitimate comes close.

## Two more found while testing

- **`getimagesize()` was called twice**, and the first result indexed for `['mime']` with no check — a non-image file emitted `Trying to access array offset on false` before reaching the switch that would have rejected it. Now one guarded call feeding both the type switch and the dimension check.
- The decode is suppressed with `@`, since libpng writes its own warnings for a file we are about to reject deliberately. Without that, `phpunit.xml`'s `failOnWarning` would fail the suite on the very case being tested.

## Testing

5 tests using a genuinely truncated PNG and a header-patched oversized one, not mocks.

| Fix | Reverted → |
|---|---|
| Decode guard | 1 test fails |
| Dimension cap | 1 test fails |
| `Throwable` widening | **0 tests fail** |

That last row is stated rather than papered over: with the decode guard in place the throw is already a `RuntimeException`, which `catch (\Exception)` handles — so the widening is defence in depth against future Errors from GD, not something these tests exercise. It costs nothing and I have kept it, but it is not covered.

Full suite **1533 tests / 6057 assertions**; test images cleaned up, no residue in `images/`.

### DIFF
--- a/admin/src/Helper/Cwmimagelib.php
+++ b/admin/src/Helper/Cwmimagelib.php
@@ -25,6 +25,16 @@ namespace CWM\Component\Proclaim\Administrator\Helper;
 class Cwmimagelib
 {
     /**
+     * Largest source edge, in pixels, that will be decoded.
+     *
+     * Output is a 300x169 thumbnail, so anything approaching this is already
+     * far larger than needed.
+     *
+     * @since  __DEPLOY_VERSION__
+     */
+    public const int MAX_SOURCE_DIMENSION = 8000;
+
+    /**
      * Extension Name
      *
      * @var string
@@ -78,8 +88,11 @@ class Cwmimagelib
                 self::resizeImage($newFilePath, $origFilePath);
 
                 return $newFileName;
-            } catch (\Exception $e) {
-                // Fall back to original if resize fails
+            } catch (\Throwable $e) {
+                // Fall back to the original image. Throwable rather than
+                // Exception because GD raises TypeError -- an Error -- when
+                // handed a failed decode, and that would otherwise escape and
+                // take down every page rendering this list.
                 return $img;
             }
         }
@@ -109,8 +122,14 @@ class Cwmimagelib
         float $canv_width = 300,
         float $canv_height = 169
     ): void {
-        $info = getimagesize($originalFile);
-        $mime = $info['mime'];
+        $info = @getimagesize($originalFile);
+
+        if (!\is_array($info) || (int) ($info[0] ?? 0) < 1 || (int) ($info[1] ?? 0) < 1) {
+            throw new \RuntimeException('Could not read image dimensions from ' . $originalFile);
+        }
+
+        [$width, $height] = [(int) $info[0], (int) $info[1]];
+        $mime             = (string) ($info['mime'] ?? '');
 
         switch ($mime) {
             case 'image/jpeg':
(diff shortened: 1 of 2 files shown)


bcordis:

PR #1629, opened 2026-08-07
fix: make episode numbers unique within a series

### LINKED ISSUE #1579: CwmepisodenumberHelper: nextNumber()/findDuplicate() TOCTOU race lets concurrent saves create duplicate episode numbers — same class #1505's audit tooling detects after the fact

## `nextNumber()` + `findDuplicate()` TOCTOU race lets concurrent saves create duplicate episode numbers

`admin/src/Helper/CwmepisodenumberHelper.php:43-58,73-88`. `CwmmessageModel::prepareTable()` calls `nextNumber()` to auto-fill a blank `studynumber` with `MAX(existing)+1`, and `CwmmessageTable::check()` later calls `findDuplicate()` to reject a collision — both plain SELECTs with no locking (`FOR UPDATE`) or transaction, and the schema has no unique index on `(series_id, studynumber)`.

Two admins adding new messages to the same series near-simultaneously, both leaving Episode Number blank: A's `nextNumber()` reads `MAX=4` → sets `5`; B's `nextNumber()` also reads `MAX=4` (A's row isn't inserted yet) → sets `5`; A's `findDuplicate(seriesId, '5')` finds none (A hasn't stored yet) → passes; B's `findDuplicate()` also finds none → passes; both `store()` succeed → two messages with `studynumber='5'` in the same series. This is exactly the class of duplicate the project built `CwmepisodeauditModel`/`findAllDuplicates()` to detect *after the fact* (per the `check()` comment referencing #1505) — narrowing the window doesn't close it.

**Fix direction:** Either add a unique index on `(series_id, studynumber)` (with a migration to first resolve existing duplicates) and let the INSERT fail atomically, or wrap the read-check-write in a transaction using `SELECT ... FOR UPDATE` / an advisory lock per series.

bcordis:
nextNumber() read MAX(studynumber)+1 and findDuplicate() checked for a clash,
both plain SELECTs with no lock and nothing in the schema behind them. Two
administrators adding a message to the same series with the number left blank
both read the same maximum, both passed the duplicate check because neither had
stored yet, and both saved. That is the duplicate the episode audit tooling was
built to find after the fact; narrowing the window never closed it.

The database now enforces it, through a generated column rather than a plain
unique index. A plain index was not available: studynumber defaults to '' and
MySQL treats '' as a value rather than as absent, so 685 of 827 unnumbered rows
on a development database would have collided with each other. Episode numbers
also only mean anything inside a real series, and series_id 0 and -1 are the
seriesless sentinels -- most existing duplicates were there, which is correct
data rather than a fault. Mapping both cases to NULL leaves them unconstrained,
because NULLs are distinct from one another in a unique index. The expression
matches findAllDuplicates()'s own filters exactly, so the constraint and the
audit tooling agree on what a duplicate is.

The migration blanks the later members of each existing duplicate group before
adding the index, since ADD UNIQUE fails outright while any remain. Blanking
rather than renumbering: those messages return to the unnumbered state the UI
already handles and an administrator renumbers them deliberately from the audit
screen, rather than the database inventing an order that a number-sorted
podcast feed would then follow. Rehearsed against a copy of the real table --
eight duplicate groups to zero, fourteen rows blanked, 641 seriesless rows
untouched -- then applied to the development database with the same result.

A constraint turns the losing side of a race into a failed write, so
CwmmessageTable::store() translates that into the same kind of message the
pre-check produces. It matches on the index name, so an unrelated duplicate-key
error is not reported as an episode-number clash.

Three existing test files seeded duplicate groups as fixtures, which the
database now refuses. Two of them did not need a real group at all -- they
needed a number to be *taken* -- so their fixtures drop to one holder per
number and every behaviour they cover is intact. The third, and
findAllDuplicates()'s own test, did need one, and that state is no longer
constructible: both now assert that a series with distinct numbers is not
reported. That is a genuine reduction in coverage of the audit tooling's
positive case, recorded here rather than hidden, and it is the direct cost of
making the state it detects impossible to create.

Fixes #1579

---

**Verification.** Suite green locally on PHP 8.5.7; each behavioural change red-checked by reverting it and confirming the relevant tests fail. Branch merges cleanly into `development`.


bcordis:
Rebased onto `development` after #1620 and #1616 merged. All three branches added `admin/sql/updates/mysql/10.5.6-20260806.sql`; the three sets of statements now live in that one file, in merge order — #1574 (topic uniqueness), #1611 (session hash), then #1579 (episode numbers).

Rehearsed all three together against copies of the real tables: 68 duplicate topic pairs to zero, both `uq_study_topic` and `uq_series_studynumber` created, 641 seriesless rows untouched, no errors.

⚠️ One honest gap in that rehearsal: the episode de-duplication step ran against zero duplicate groups, because this development database had already had them blanked during earlier verification. That path was rehearsed separately when the migration was written — eight groups to zero, fourteen rows blanked — but not in this combined run.

Commits re-signed after the rebase.

Issue #1662, opened 2026-08-08
Consumer registry keeps rows for extensions removed by a package uninstall


bcordis:
Observed while tracing lib_cwmscripture#37.

## What happens

After `pkg_proclaim` is removed, `#__bsms_scripture_consumers` still lists both of the extensions that were just uninstalled:

```
consumers registry after removal:
  component          com_proclaim
  plugin    content  scripturelinks
```

Neither exists in `#__extensions` any more, and neither is on disk.

## Why

`ConsumerRegistry::installedExcluding()` does prune stale rows — that part works:

```php
if (!self::isInstalled($db, $consumer)) {
    // Registered but no longer present — drop the stale row.
    if (!\in_array($consumer, self::FIRST_PARTY, true)) {
        self::unregister($consumer['element'], $consumer['type'], $consumer['folder']);
    }
    continue;
}
```

The problem is *when* it runs. `InstallerAdapter::uninstall()` calls `triggerManifestScript('uninstall')` — where the package's cleanup lives — **before** `removeExtensionFiles()`, which is what removes the children. So at the moment the pruning happens, `com_proclaim` and `scripturelinks` are still installed and are correctly left alone. By the time they are gone, nothing runs again.

The ordering is right for the table decision (the package needs the registry readable and the children still present, which is why it excludes them explicitly). It is simply one pass too early to also collect the corpses.

## Effect

Low. The rows are inert while nothing is installed, and `installedExcluding()` prunes them the next time it runs — which in practice is the next uninstall after a reinstall. But:

- a reinstall inherits a registry describing a previous installation
- anything reading the registry directly, rather than through `installedExcluding()`, sees extensions that do not exist
- it makes the registry a poor diagnostic, which is what sent lib_cwmscripture#37 down the wrong path initially

## Possible directions

1. Have each consumer unregister itself in its own uninstall script — accurate, but spreads the responsibility across three repos.
2. Have the package's `uninstall()` explicitly unregister the members it is about to remove, since it already names them for the exclusion list. One place, and the list is right there.
3. Accept it and rely on the next `installedExcluding()` to tidy up — then document that the registry can name uninstalled extensions, so nobody trusts it raw.

Option 2 looks cheapest: `dropScriptureTablesIfLastConsumer()` already has the exact array.

Issue #1661, opened 2026-08-08
Package uninstall leaves libraries/cwmscripture on disk after its extension row is gone


bcordis:
Observed while running `test-upgrade.sh` step 11 by hand against j6-test (see lib_cwmscripture#37 for the full trace).

## What happens

After `extension:remove` on `pkg_proclaim`, the library's row is gone from `#__extensions` but **its files remain on disk**:

```
administrator/components/com_proclaim          gone
plugins/content/scripturelinks                 gone
administrator/manifests/packages/proclaim      gone
libraries/cwmscripture                         PRESENT   <--
```

The package removal reported `[OK] Extension removed!` and every other child was cleaned up correctly, so this is specific to the library.

## Why it matters

A later reinstall lays new files over an old tree rather than a clean one. That is the shape of bug that leaves a stale class behind and only surfaces much later — the kind #25 in lib_cwmscripture already cost us once, where minified output survived a source rename because nothing removed it.

It is also a plain leak: an administrator who removes Proclaim entirely still has `libraries/cwmscripture/` on the server.

## Not yet investigated

Whether this is Joomla's `LibraryAdapter::removeExtensionFiles()` declining to remove the folder, or something in the library's own `script.php::uninstall()` returning early before file removal. `dropTablesIfOrphaned()` returns early on `isPackageUninstall()`, but that guard is about tables — worth checking whether an early return there skips anything else.

Reproduction is one command on a site with pkg_proclaim installed:

```bash
php cli/joomla.php extension:remove -n <pkg_proclaim extension_id>
ls -d libraries/cwmscripture
```

bcordis:
True fact, Proclaim will never uninstall the scripture or scripture library using the package UnInstaller as many times it may be being used elsewhere. So this may be a mute problem, but we may just need to have a notice saying that the following extra packages/components/plug-ins are still installed.

bcordis:
Root cause found, and it turns out to be **neither** of the two candidates in the description.

`<files>` ships directories only — `src`, `sql`, `language` — while Joomla copies `<scriptfile>` into the same extension root from the **package** root, where `<files>` cannot reference it. `LibraryAdapter::removeExtensionFiles()` clears the three folders, then skips the delete because the empty-check still sees one file:

```php
$files = Folder::files($this->parent->getPath('extension_root'));
if (!count($files)) { Folder::delete(...); }
```

[...]

PR #1663, opened 2026-08-08
test: make step 11 check its own precondition (lib_cwmscripture#37)

bcordis:
Resolves the harness side of lib_cwmscripture#37, which turned out to be a **false alarm** — the library and `pkg_proclaim`'s cleanup both behaved correctly.

## What was wrong

Step 11 asserted that removing `pkg_proclaim` drops the scripture tables. On j6-test, `com_livingword` is installed — it's a `FIRST_PARTY` consumer and genuinely uses those tables (`LivingwordComponent.php` references `CWM\Library\Scripture`). Keeping them was **correct**; dropping them would have broken Living Word.

The phase was asserting an outcome its own preconditions ruled out, and had no way to report that. Confirmed by probing `ConsumerRegistry` directly against the live site:

```
installedExcluding() => ["Living Word (com_livingword)", "com_livingword (detected)"]  -> would KEEP
```

## Three changes, none to the library

**`assert-no-other-consumer`** — new mode, run *before* the removal. Afterwards the registry rows and extension rows are gone and the question is unanswerable.

The blocker query already existed, inside `assert-detected-consumer` (step 13), which has had this guard all along and documents exactly why: *"A vacuous pass is worse than no test."* Step 11 simply never got one. It's now a shared closure used by both, so the duplication goes too.

**`assert-tables-gone` explains itself.** Four bare `FAIL` lines look the same whether a consumer legitimately kept the tables or the cleanup never ran — and the first was mistaken for the second. It now names the responsible consumer, or, when there is none, points at `pkg_proclaim`'s `script.install.php::uninstall()` and notes that it logs to the `jerror` category, which has **no logger under CLI** — so silence there is not evidence the cleanup was skipped. That red herring cost real time.

**The `|| true` is gone.** A removal that failed and one that succeeded then correctly declined are indistinguishable downstream. Step 10 immediately above already asserts a *non-zero* exit for the standalone library, so the status was clearly understood to matter here too.

## Verified live

Against j6-test with `com_livingword` installed:

```
=== assert-no-other-consumer ===
  FAIL another scripture consumer is installed: com_livingword (first-party, installed).
       Removing pkg_proclaim will correctly KEEP the tables, so asserting
       they are dropped would be asserting a bug.

=== assert-tables-gone ===
  FAIL j6test_bsms_bible_translations still exists after the last consumer was removed.
  ...
       Reason: another consumer is installed — com_livingword (first-party, installed).
       That makes keeping the tables CORRECT.
```

`composer lint` clean (0 of 613); `bash -n` clean.

## Follow-ups filed

Two unrelated observations from the same run, both low priority: #1661 (`libraries/cwmscripture` left on disk after package removal) and #1662 (consumer registry keeps rows for removed extensions).

### DIFF
--- a/build/test-upgrade.sh
+++ b/build/test-upgrade.sh
@@ -103,6 +103,37 @@ if [ ! -f "$JCLI" ]; then
     exit 1
 fi
 
+# Phase 11 asserts that removing pkg_proclaim drops the shared scripture tables,
+# which is only true when nothing else uses them. Any other consumer -- Living
+# Word is the one that actually bit us (lib_cwmscripture#37) -- makes keeping
+# them correct, so the assertion would be testing for a bug.
+#
+# Clear them here rather than in phase 11 itself: phases 12 and 13 install their
+# own consumers to prove the opposite case, and an unrelated one already present
+# would let those pass without their fixtures doing anything.
+#
+# NOTE: reset-testsite.php only removes the Proclaim family, so anything taken
+# out here stays out. That is fine for a role=test site -- everything from this
+# point is destructive by design -- but it does mean a consumer must be
+# reinstalled by hand if a later run needs it.
+# grep to digits only: this mode prints bare ids with no header, so a stray PHP
+# notice on stdout would otherwise be passed to extension:remove as an id.
+OTHER_CONSUMERS="$(php build/verify-scripture-uninstall.php other-consumer-ids | grep -E '^[0-9]+$' || true)"
+
+if [ -n "$OTHER_CONSUMERS" ]; then
+    echo "   clearing other scripture consumers before the uninstall phases:"
+
+    for CID in $OTHER_CONSUMERS; do
+        if php "$JCLI" extension:remove -n "$CID" >/dev/null 2>&1; then
+            echo "     removed extension id ${CID}"
+        else
+            echo "ERROR: could not remove scripture consumer id ${CID}." >&2
+            echo "       Phase 11 would then fail for the wrong reason." >&2
+            exit 1
+        fi
+    done
+fi
+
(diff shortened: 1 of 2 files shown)


bcordis:
## Proven end to end — and #37 is confirmed a false alarm

Second commit adds the missing half: the uninstall-phase preamble now clears any scripture consumer that is not part of `pkg_proclaim`, so phase 11's precondition can actually be met rather than merely detected.

`other-consumer-ids` is deliberately **not** hardcoded to `com_livingword` — assuming which consumers a test site carries is what broke this in the first place. It resolves `FIRST_PARTY` members and registered third parties alike, and prefers a parent package over its component so a removal cannot leave a package owning nothing. Each removal's exit status is checked, since a consumer that failed to uninstall would otherwise surface as a phase 11 failure with no visible connection to the cause.

## Run on j6-test with the field clear

```
=== [precondition] ===
  OK   no other scripture consumer — a package removal should drop the tables

=== [removal] pkg id 1782 ===
 [OK] Extension removed!

=== [assertion] ===
  OK   j6test_bsms_bible_translations removed — nothing was left using it
  OK   j6test_bsms_bible_verses removed — nothing was left using it
  OK   j6test_bsms_scripture_cache removed — nothing was left using it
  OK   j6test_bsms_scripture_consumers removed — nothing was left using it
```

**Phase 11 passes.** `lib_cwmscripture`'s guard and `pkg_proclaim`'s `script.install.php::uninstall()` were both working correctly the whole time — the tables were kept because Living Word genuinely uses them, which is the guard doing its job.

Nothing in the library needs changing, and lib_cwmscripture#37 can be closed as invalid.

`bash -n` clean, `composer lint` clean (0 of 613).

Issue #1653, opened 2026-08-08
Make Proclaim's per-section permissions real, so a manager or custom group can be limited to parts of the component


bcordis:
Raised by @bcordis while reviewing the control-panel tile gating in #1652: *"Right now it's not possible to set permissions to each section of proclaim… we may need to make a plan to move to that so a manager/new group can get limited access to sections of Proclaim."*

## Where we actually are

`admin/access.xml` **already declares** 17 sections — `message`, `teacher`, `location`, `playlist`, `serie`, `topic`, `podcast`, `share`, `comment`, `mediafile`, `messagetype`, `server`, `template`, `cssedit`, `style`, `templatecode`, `admin` — each with create / delete / edit / edit.state / edit.own.

But **no section assets exist**. On a development database `#__assets` holds exactly `com_proclaim` plus per-item rows. `Access::getAssetRules()` therefore falls back to the extension asset for any `com_proclaim.<section>` lookup. Verified empirically:

```
com_proclaim             edit=true
com_proclaim.teacher     edit=true
com_proclaim.nonsense    edit=true      <- no such section
[non-super user]         edit=false for both
```

So today every section check silently resolves to the component's own rules. The declarations are inert, and there is no UI that would let an administrator set them even if they were not.

## A real bug this uncovers

`CwmserverTable::_getAssetName()` writes `com_proclaim.server.<id>`, but `CwmserverModel` authorises against **`com_proclaim.cwmserver.<id>`** (lines 387, 426, 447) — a name nothing ever creates. It currently falls back to component rules, so it appears to work.

The moment section assets exist, everything else starts honouring section rules while servers quietly keep ignoring them. **Any name mismatch has to be found and fixed before phase 1, not after** — while all names are still equivalent and a mistake is harmless.

### Phase 0 — audit the names (prerequisite)

Every `authorise()` call and every `_getAssetName()` / `_getAssetParentId()` must agree with `access.xml`. Real `authorise()` calls today use `message`, `teacher`, `serie`, `comment`, `mediafile` (all valid) plus the `cwmserver` outlier above. Also decide what to do with `share`, `cssedit` and `style`, which are declared but referenced nowhere — either wire them up or drop them from `access.xml` rather than shipping settable permissions that govern nothing.

Cheap, behaviour-neutral, and it makes every later phase safe.

### Phase 1 — create the section assets (no behaviour change)

[...]

PR #1672, opened 2026-08-09
feat: add a Permissions tab with a grid per section (#1653 phase 2)

bcordis:
> **Draft, stacked on #1671.** ⚠️ **The save path is not yet verified end to end** — see "What is not proven" below. Do not merge on the strength of the rendering evidence alone.

Phases 1 and 1.5 created the section assets and made the code ask about them. This is the UI that lets an administrator set them.

## What it adds

- `CwmadminModel::addSectionPermissionFields()` — one hidden `asset_id_<section>` and one `rules_<section>` grid per section **declared in access.xml**, generated rather than listed in `admin.xml`, so access.xml stays the single source of truth
- `CwmadminModel::saveSectionRules()` — writes each grid to its own asset, following com_config's `ComponentModel`: build a `Rules`, store it, and **refuse without `core.admin`**
- a Permissions tab on the Administration screen, gated on `core.admin`
- 20 language keys

Each grid needs its own `asset_field`, because `RulesField` reads the id from the field named there — a shared `asset_id` would point all 17 grids at the same asset.

## Two bugs found by driving the real admin

Neither was visible from reading the code, and both would have shipped.

**1. The asset ids were all empty.** They must be bound in `addSectionPermissionFields()`, not `loadFormData()`. The fields are added *after* `loadForm()` has already bound its data, so values set on the data object never reach them.

The consequence is not cosmetic: an empty `asset_field` makes `RulesField` **fall back to the component asset**. All 17 grids would have displayed — and saved — `com_proclaim`'s own rules under a section label. Confirmed in the browser: 17 fields, all `value=""`. After the fix, 17 distinct ids (3417, 3418, 3419…).

**2. A nested tab set does not work here.** The outer `myTab` set never emits an `endTabSet`, so a tab set opened inside one of its tabs renders panels that never become visible, leaving the permission selects unreachable. The sections are now stacked `<details>` blocks instead.

## What is verified

In a real browser on j6-dev, logged in as an administrator:

```
joomla-field-permissions grids : 17
perm-<section> panels          : 17
hidden asset_id inputs         : 17  (all populated, all distinct)
sections rendered              : admin, comment, cssedit, location, mediafile,
                                 message, messagetype, playlist, podcast, serie,
                                 server, share, style, teacher, template,
                                 templatecode, topic
PHP errors on page             : none
```

`composer test:unit` — 1399 tests, 2995 assertions, OK. `php-cs-fixer` clean.

## What is NOT proven

**Saving a section rule.** A submit driven with `Joomla.submitbutton('cwmadmin.apply')` reloaded the page without reaching `CwmadminModel::save()` — instrumentation in `saveSectionRules()` never fired, and `com_proclaim.teacher` stayed `{}`. So the write path is written and reviewed but unexercised.

That could be the harness rather than the code — the grids nest their own `joomla-tab` per user group, so driving them properly means activating a group tab before a select is reachable, and my shortcut of setting `.value` directly may simply not have submitted. But it is unproven either way, and the correct end-to-end check is the one from #1653: set *Teachers → Edit → Denied* for Manager, save, and confirm a Manager loses the Teachers toolbar while keeping Series.

## Note on j6-dev

Loading the tab created the 17 section assets there, since `sectionId()` finds-or-creates. That is phase 1 behaviour working as intended, not a side effect of this PR, but it means j6-dev now reflects post-phase-1 state.


bcordis:
Superseded by #1717, which replaces the tab with a dedicated `view=cwmpermissions` screen.

The save path here could never be verified, and the cause was not the harness. This form posted **1217 fields** against PHP's default `max_input_vars` of 1000:

```
PHP Warning:  PHP Request Startup: Input variables exceeded 1000.
```

PHP truncates `$_POST` at request startup with only that warning. Joomla's `task` field sits near the end of the form, so it was dropped, the request dispatched to `display`, and no controller ran — which is exactly why instrumentation in `saveSectionRules()` never fired. Nothing was wrong with the model.

Field count is `actions × user groups`, so 17 grids break at roughly twelve groups. Multi-campus sites are exactly the sites with that many. #1717 carries one grid per page — 73 fields — and its save round trip is verified against `#__assets`.

The two real bugs found here (the unbound asset ids, and the nested tab set) both informed the replacement: the new form declares its fields statically so binding happens normally, and there is no nested tab set.

Closing as superseded rather than merging.

Issue #1407, opened 2026-08-01
Cache clearing on update: duplicated plugin name, and a folder-delete warning


bcordis:
Both surfaced on a real 10.5.0 update, in the cache-clearing feature added by #1378. Neither is fatal — the update completes — but both are visible to every administrator on every update.

## 1. The external-cache notice repeats the plugin name

Observed:

> Proclaim cleared Joomla's cache, but **JCH Optimize, JCH Optimize** manages its own and Proclaim cannot clear it.

`warnAboutExternalCaches()` in `proclaim.script.php` queries `#__extensions` filtered by `type = 'plugin'` and `enabled = 1`, but **not by `folder`**:

```php
->select($db->quoteName('element'))
->from($db->quoteName('#__extensions'))
->where($db->quoteName('type') . ' = ' . $db->quote('plugin'))
->where($db->quoteName('enabled') . ' = 1')
->whereIn($db->quoteName('element'), array_keys($known), ParameterType::STRING);
```

JCH Optimize registers more than one plugin under the element `jchoptimize`, so `loadColumn()` returns it twice, both rows map to the same display name, and `implode(', ', $found)` prints it twice.

**Fix:** `array_unique()` on the mapped names. Filtering by folder would also work but is more brittle — any one of the plugins being enabled is reason enough to warn, and the folder names vary by release.

The warning is otherwise doing its job correctly: it identified a real cache Proclaim cannot clear and told the administrator to clear it.

## 2. Clearing the cache raises a folder-delete warning

Observed:

> Warning — Could not delete folder. Path: `[ROOT]/administrator/cache/language`

That comes from Joomla itself, `joomla/filesystem/src/Folder.php:270`, thrown after `@rmdir()` fails:

```php
throw new FilesystemException(sprintf('%1$s: Could not delete folder. Path: %2$s', __METHOD__, $path));
```

`clearCaches()` calls `Factory::getCache('')->clean()` with no group — the equivalent of com_cache's *Delete all* — which walks into the language cache directory Joomla is using during the update itself.

The method's own docblock states the intent: *"Never fatal: an update must not fail over cache hygiene."* It isn't fatal, but it does produce a red warning that looks like the update went wrong when it didn't.

**Fix options:**

- Clean the groups that matter rather than everything, leaving the language cache to Joomla.
[...]

PR #1683, opened 2026-08-10
fix: split the compound ALTERs Joomla only half-checks, and reconcile uq_aggregate in PHP (#1664)

### LINKED ISSUE #1664: Analytics monthly aggregate: unguarded DROP INDEX, and a compound ALTER that Joomla's schema check silently skips
Spotted while fixing the equivalent problem in `10.5.6-20260807.sql` (see the `dropRedundantStudyTopicIndex()` change shipping in 10.5.6). That fix deliberately stopped at the unreleased migration; this one is in a long-released file and wants its own decision.

## The statement

`admin/sql/updates/mysql/10.1.0-20260225.sql:11-18`

```sql
ALTER TABLE `#__bsms_analytics_monthly`
    ADD COLUMN `series_id` INT UNSIGNED NULL DEFAULT NULL AFTER `study_id`,
    DROP INDEX `uq_aggregate`,
    ADD UNIQUE KEY `uq_aggregate` (
        `series_id`, `study_id`, `media_id`, `location_id`,
        `event_type`, `referrer_type`, `country_code`, `device_type`,
        `year`, `month`
    );
```

Two separate problems, and the second is the interesting one.

## 1. The `DROP INDEX` is unguarded

MySQL has no `DROP INDEX IF EXISTS`. A database where `uq_aggregate` is absent stops this statement with **error 1091**, and because `Installer::parseSchemaUpdates()` executes update SQL raw, that aborts the whole update.

Same root cause as the 10.5.6 case: reachable when the tables and the recorded `#__schemas` version disagree. `Cwmrestore::restore()` clears `#__schemas` (`admin/src/Lib/Cwmrestore.php:218`) precisely so migrations re-run, so restored databases are the realistic route in.

## 2. The compound ALTER defeats Joomla's replay check — this is the worse half

`MysqlChangeItem::buildCheckQuery()` decides what a statement does from **`$wordArray[3] . ' ' . $wordArray[4]`** only (`libraries/src/Schema/ChangeItem/MysqlChangeItem.php:88`). For this statement that reads `ADD COLUMN`, so the check query it builds is:

```sql
SHOW COLUMNS IN `#__bsms_analytics_monthly` WHERE field = 'series_id'
```

The index rework is invisible to it. So on the restore path (`Cwmrestore` → `DatabaseModel::fix()` → `ChangeSet`):

- `series_id` exists → `checkStatus = 1` → "already applied"
- `ChangeItem::fix()` only executes when `checkStatus === -2` (`ChangeItem.php:225`)
- **the `DROP INDEX` / `ADD UNIQUE KEY` half never runs**

A database that has `series_id` but still carries the *old* `uq_aggregate` therefore keeps the wrong unique key forever, and no amount of re-running the schema fix will correct it. The uniqueness constraint for analytics aggregates is then narrower than intended — rows that should collide into one monthly aggregate can co-exist, which is a silent double-counting shape rather than an error anyone sees.

## Why it was not fixed alongside the 10.5.6 one

Editing a released migration changes what already-updated sites would replay. That needs deciding, not assuming — hence this issue rather than a drive-by change.

## Options

1. **Split the compound ALTER into single-action statements.** Each becomes independently parseable, so `ChangeSet` can check and repair the index on its own. Does not by itself fix the 1091, but makes the statement honest to Joomla's mechanism.
2. **Move the index rework into the install script**, mirroring `dropRedundantStudyTopicIndex()` — read `SHOW INDEX`, drop only if present, re-add if its column list is wrong. Fixes both halves, and repairs databases already carrying the wrong key.
3. **Leave the released file alone and add a forward repair** in the next version's migration/postflight. Safest for sites that already ran it, and the only option that actually corrects existing wrong keys without touching history.

Option 3 plus 2's repair logic looks like the right combination: do not rewrite what shipped, add a guarded reconcile that runs going forward.

## Verifying

Worth measuring before choosing: on a dev database, compare `SHOW INDEX FROM #__bsms_analytics_monthly` against the 10-column list above. If real sites are already carrying the pre-10.1.0 key, that settles it as a repair job rather than a hardening job.

## Not a 10.5.6 blocker

The normal upgrade path is unaffected — sites that ran 10.1.0 in sequence had `uq_aggregate` present and got the rework. This is about restored databases and replayed schemas.

bcordis:
Closes #1664 — with more in it than the issue describes, and one severity claim corrected.

## The mechanism

`MysqlChangeItem::buildCheckQuery()` decides what an `ALTER` does from **words three and four** of the statement and nothing else. A multi-clause ALTER is represented by its first clause alone: every later clause is never verified, never repaired, and counted as applied whenever the first happens to be in place. On the restore path — `Cwmrestore` clears `#__schemas` so `DatabaseModel::fix()` replays every migration through `ChangeSet` — those clauses simply never run.

## A sweep found three, not one

The issue named the `uq_aggregate` statement. Scanning the whole migration set found **three compound ALTERs, all in 10.1.0**, and the one it didn't mention is the worse find:

```sql
ALTER TABLE `#__bsms_podcast`
    ADD COLUMN `itunes_category` …,      -- the only clause Joomla ever saw
    ADD COLUMN `itunes_subcategory` …,
    ADD COLUMN `itunes_explicit` …,
    ADD COLUMN `itunes_type` …;
```

A database holding `itunes_category` and nothing else reads as up to date, and no schema fix will ever add the other three. **A missing column is a SQL error on the podcast form, not a silent miscount.** The third hid `idx_series_created` behind the events `series_id` column.

Editing released migrations is safe here: a site past 10.1.0 never replays them on the update path, and a site below it runs the same clauses in the same order. Only the *check* path changes.

## Why uq_aggregate left SQL entirely rather than being split

Neither shape works:

- **Compound** — the index half is invisible, which is the bug.
- **Split** — a lone `DROP INDEX` becomes a check item with `checkQueryExpected = 0`, and the very next statement re-adds that name. It would read as unapplied on every check and be **re-dropped by every fix**: a permanently red Database view that actively destroys the key.

So it moves to `reconcileAnalyticsAggregateIndex()` in `proclaim.script.php`, which can read the state first. That also picks up a case **no SQL formulation reaches** — a key over the *wrong columns*. ChangeSet's index check only looks at the name.

Verified against a scratch table in four states — correct key (no-op), the pre-10.1.0 nine-column key (rebuilt), the right ten columns in the wrong order (rebuilt), no key at all (added).

The catch logs rather than aborts, and it is load-bearing: re-adding a unique key over violating rows throws — but **only for rows whose ten dimensions are all non-NULL and equal**, since MySQL treats NULLs as distinct. I confirmed both halves of that empirically.

## Correcting the issue, and correcting #1682

**The issue's severity claim is wrong.** It describes the wrong key as "a silent double-counting shape." It isn't: `CwmanalyticsHelper::rollup()` already deletes each matching group before inserting it, with a comment saying the key only works for entirely-non-NULL rows anyway. The real cases are the **1091 abort** (hard failure) and **unrepairable schema drift** (mechanism, not observed).

**And #1682's body overstated what the release gate does.** `verify-migrations.php` resolved a *single* `$key`, so `idx_playlist_video` in the 10.3.3 block never ran for a 10.5.x release. Fixed here: every entry at or below the version under test is now verified, oldest first, each assertion labelled with the version that introduced it.

## Measured before deciding

The issue asked for this. All five dev sites carry the correct 10-column `uq_aggregate`, `idx_series_created`, and all four podcast columns — so this is **hardening, not repair**. Worth knowing before shipping a migration that claims to fix data.

## The durable part

Fixing three statements doesn't stop a fourth. `MigrationStatementContractTest` asserts every ALTER carries exactly one clause and none drops an index — **157 statements, 314 assertions**. Both rules verified by mutation: restoring the four-clause podcast ALTER and appending a `DROP INDEX` each turn a named assertion red.

The clause counter ignores commas inside quotes and parentheses. My first pass without that reported two single-clause statements as compound, so `10.1.0-20260228.sql` (`COMMENT '0=ministry-created, 1=external/third-party'`) is now the standing control.

Registered in `phpunit.xml` — a new `tests/unit` directory with no suite entry passes by never running.

## Results

- Full suite **2134 tests, 7178 assertions**, green.
- Migration gate against j6-test: **35 schema assertions pass, 9 of them new.** The one failure, `#__schemas >= 10.5.6 (found none)`, is identical before and after this change — j6-test currently has no `com_proclaim` extension row at all, its `bsms_` tables left behind by the uninstall work exactly as designed.

## No changelog entry

Deliberate. Nothing here is observable on a healthy site, and no site was found in the broken state. Adding a line would advertise a data problem nobody has.

### DIFF
--- a/admin/sql/updates/mysql/10.1.0-20260225.sql
+++ b/admin/sql/updates/mysql/10.1.0-20260225.sql
@@ -3,16 +3,30 @@
 -- Enables tracking visits to series pages directly, and attributes all
 -- message-level events (plays, downloads, page views) to their series.
 --
+-- One clause per statement. MysqlChangeItem reads a statement's meaning from its
+-- third and fourth words only, so a compound ALTER is checked and repaired by its
+-- first clause alone; everything after it is invisible to System > Database and
+-- to the schema fix a restore runs. These were one ALTER each (#1664).
+--
 
 ALTER TABLE `#__bsms_analytics_events`
-    ADD COLUMN `series_id` INT UNSIGNED NULL DEFAULT NULL COMMENT 'FK #__bsms_series' AFTER `study_id`,
+    ADD COLUMN `series_id` INT UNSIGNED NULL DEFAULT NULL COMMENT 'FK #__bsms_series' AFTER `study_id`;
+
+ALTER TABLE `#__bsms_analytics_events`
     ADD KEY `idx_series_created` (`series_id`, `created`);
 
 ALTER TABLE `#__bsms_analytics_monthly`
-    ADD COLUMN `series_id` INT UNSIGNED NULL DEFAULT NULL AFTER `study_id`,
-    DROP INDEX `uq_aggregate`,
-    ADD UNIQUE KEY `uq_aggregate` (
-        `series_id`, `study_id`, `media_id`, `location_id`,
-        `event_type`, `referrer_type`, `country_code`, `device_type`,
-        `year`, `month`
-    );
+    ADD COLUMN `series_id` INT UNSIGNED NULL DEFAULT NULL AFTER `study_id`;
+
+-- The uq_aggregate rework that used to sit here — DROP INDEX followed by a wider
+-- ADD UNIQUE KEY over the same name — is now done by proclaim.script.php's
+-- reconcileAnalyticsAggregateIndex(). It cannot be expressed here safely:
+--
+--   * MySQL has no DROP INDEX IF EXISTS, so a database missing the key stops the
+--     statement with error 1091, and parseSchemaUpdates() aborts the whole update.
+--   * Split into its own statement, the DROP becomes a check item that expects
+--     zero matching indexes forever — and the key is re-added on the next line, so
+--     it would read as unapplied on every check and be re-dropped by every fix.
+--
+-- The PHP reconcile has neither problem, and unlike any SQL formulation it can
+-- also repair a key that exists with the wrong columns.
--- a/admin/sql/updates/mysql/10.1.0-20260304.sql
+++ b/admin/sql/updates/mysql/10.1.0-20260304.sql
@@ -1,6 +1,17 @@
--- Add iTunes configuration columns to podcast table
+-- Add iTunes configuration columns to podcast table.
+--
+-- One ALTER per column. As a single four-clause statement, Joomla's schema check
+-- built its check query from the first clause only, so itunes_category was
+-- verified and the other three were invisible — a database missing them would be
+-- reported as up to date, and the schema fix would never add them (#1664).
 ALTER TABLE `#__bsms_podcast`
-    ADD COLUMN `itunes_category` VARCHAR(100) NOT NULL DEFAULT 'Religion & Spirituality' AFTER `linktype`,
-    ADD COLUMN `itunes_subcategory` VARCHAR(100) NOT NULL DEFAULT 'Christianity' AFTER `itunes_category`,
-    ADD COLUMN `itunes_explicit` VARCHAR(5) NOT NULL DEFAULT 'false' AFTER `itunes_subcategory`,
+    ADD COLUMN `itunes_category` VARCHAR(100) NOT NULL DEFAULT 'Religion & Spirituality' AFTER `linktype`;
+
+ALTER TABLE `#__bsms_podcast`
+    ADD COLUMN `itunes_subcategory` VARCHAR(100) NOT NULL DEFAULT 'Christianity' AFTER `itunes_category`;
+
+ALTER TABLE `#__bsms_podcast`
+    ADD COLUMN `itunes_explicit` VARCHAR(5) NOT NULL DEFAULT 'false' AFTER `itunes_subcategory`;
+
+ALTER TABLE `#__bsms_podcast`
     ADD COLUMN `itunes_type` VARCHAR(10) NOT NULL DEFAULT 'episodic' AFTER `itunes_explicit`;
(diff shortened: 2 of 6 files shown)


bcordis:
Three follow-ups pushed after review.

**The catch could not report.** `Log::add(..., 'jerror')` routes into the message queue on a web request — that is how #1407's cache warning reached the screen — but under CLI nothing renders the queue and no logger is registered for the category. `composer test:install` and `test:upgrade` both run postflight headless, so on the path where the message matters most it went nowhere. Now two channels: `enqueueMessage` for the browser, `CwmlogHelper`'s self-registering file logger for CLI. Safe to depend on the component's helper there — the method returns early when the table is absent, so it only ever does work on an update.

**Confirmed the ordering the split depends on.** `10.1.0-20260225.sql` no longer touches the index at all, so a sub-10.1.0 upgrade gets `series_id` from SQL and the key from postflight. That is only correct if postflight runs after the schema updates *in the same request* — `InstallerAdapter` calls `parseQueries()` before `triggerManifestScript('postflight')` on all three paths (lines 410/445, 767/802, 1212/1249), install and update alike. Checked rather than assumed.

**The `#__schemas` failure was a misleading message, not schema drift.** j6-test has no `com_proclaim` row at all — its `bsms_` tables were left behind by the uninstall work behaving exactly as designed — and Proclaim's tables outliving Proclaim means every table/column/index assertion still passes on such a site. The only symptom was a version mismatch that was never the problem. The gate now says so and names the fix.

That last one matters for release, so recording it here rather than leaving it in a PR body: **the release gate reads red on j6-test until Proclaim is reinstalled there.** `composer test:release` installs before verifying, so a real release run is unaffected — but anyone running `verify-migrations.php` standalone will now get an accurate reason instead of a puzzle.

Also verified the data provider's own guard: pointing it at a missing directory produces provider errors and `No ALTER TABLE statements found — is the updates directory still there?`, not a silent pass.

Full suite still green at 2134 tests / 7178 assertions.

PR #1685, opened 2026-08-10
fix: stop the E2E API job timing out, and make a failing nightly reach a human

bcordis:
You're right that it's happening a lot: **three failures in the last fifteen E2E runs** — 2026-08-07, 2026-08-08, and the first push of #1664 — all the same shape.

```
page.goto: Timeout 30000ms exceeded
navigating to "http://127.0.0.1:8890/administrator/index.php", waiting until "networkidle"
```

Always a retry away from green, which is why it read as noise. It isn't. Two things in the setup guaranteed this would happen eventually.

## 1. The worker pool was smaller than one page's connections

`networkidle` resolves after **500ms with no request in flight**. Chromium opens up to **six connections per host**. The dev server was started with **four workers**:

```
PHP_CLI_SERVER_WORKERS=4 php -S 127.0.0.1:8890 …
```

So two requests out of any six sat queued behind whichever worker was slowest, and the idle window couldn't open until they drained. Now 10 — above what a single page can ask for.

## 2. What kept the workers busy was an internet round trip

The dashboard's update quickicons fire an AJAX that **core itself introduces with this comment**:

```php
// Close the session before we make a long running request
$app->getSession()->abort();
```

`com_installer`'s `update.ajax` polls **every enabled update site** — and after `test:install` runs, that includes Proclaim's own on christianwebministries.org. Every CI run was taking a dependency on an external server, for a number **no test asserts on**.

Those plugins and Joomla's statistics reporting are now disabled for the disposable site, alongside `guidedtours`, which was switched off for exactly this class of reason.

## The wait strategy

The two navigations that actually timed out drop `networkidle` for `domcontentloaded`. Both are immediately followed by a wait for the specific element the navigation exists to reach — stricter than "the page stopped talking", and what Playwright recommends instead of `networkidle`.

**That change is not free in `global-setup`,** and this is the part worth reviewing. `isVisible()` answers for the instant it is called; `networkidle` was what made that safe by having already waited for everything. Under `domcontentloaded` it would be asking an empty document and getting *"no login form"* from a page that simply hadn't rendered — the same false negative the file's own comment records as having once **saved a guest session as valid auth state**. It now races the login form against the admin sidebar before deciding, so a slow render reads as "not yet".

## Scope

The other 24 `networkidle` uses are in the a11y, navigation and dragging specs — which **this job does not run** (`--project=api-test` only), and where waiting for a settled page is closer to the point. Left alone deliberately rather than swept up.

I also didn't reach for `retries: 1`. It would have hidden this rather than fixed it, and the suite has been telling us something true for a month.

## Verification

The mechanism is proven from core's source rather than by reproduction — I can't reproduce a timing-dependent external-network flake locally on demand. What the next few runs of this job show is the real test, and the scheduled nightly gives us one a day.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -138,6 +138,19 @@ jobs:
           # showed the tour modal. No tours for a disposable site.
           mysql -h 127.0.0.1 -u root -pjoomla_ci joomla_ci \
             -e "UPDATE jos_extensions SET enabled = 0 WHERE folder = 'system' AND element = 'guidedtours';"
+          # The admin dashboard's update quickicons call out to update.joomla.org
+          # and to every registered update site — which, once test:install has
+          # run, includes Proclaim's own on christianwebministries.org. Each of
+          # those AJAX calls holds a dev-server worker for the length of an
+          # internet round trip, so CI runs took a dependency on servers it does
+          # not control and gets nothing in return: no test asserts on an update
+          # count. Statistics reporting goes the same way — a disposable site
+          # should not phone home either.
+          mysql -h 127.0.0.1 -u root -pjoomla_ci joomla_ci -e "
+            UPDATE jos_extensions SET enabled = 0
+             WHERE (folder = 'quickicon'
+                    AND element IN ('joomlaupdate','extensionupdate','autoupdate','downloadkey','eos'))
+                OR (folder = 'system' AND element = 'stats');"
 
       - name: Write build.properties for the disposable site
         # The harness discovers the test site by role, not by name (#1353),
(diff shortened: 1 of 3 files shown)


bcordis:
## Second commit: the monitoring, which answers why nobody saw this

Asked why three nightly failures never surfaced. They never surfaced because **nothing reported them**.

- The only thing a scheduled failure produced was two artifacts on a page nobody opens.
- **E2E is not a required check** on `development` — the required list is `PHP Lint & Tests`, `JS Lint & Tests`, `PHP Lint & Tests (8.5)` and nothing else.
- On PR runs it *is* visible, which is how this one got noticed — but on #1664 the first run failed and the second passed, and `gh pr checks` reports the latest result per job, so the red was already gone by the time I looked.

So the scheduled run is the one with no audience, and only it opens an issue now: **one** issue, commented on each further failure rather than one per night, closed automatically when the nightly goes green. Same shape as `vendor-check.yml`, so there's one pattern here rather than two.

### Should E2E be a required check?

**No — and I'd keep it that way.** Its path filter means it doesn't report at all on most PRs, and a required check that never reports blocks them forever. That's why the issue body says so explicitly: a red nightly is information, not a merge gate.

### Two review points

**Found by label, not title.** `--search "in:title …"` is fuzzy and reads GitHub's search index, which lags creation by seconds to minutes — a re-run or manual dispatch inside that window would each find nothing and open a duplicate. `--label` hits the API directly and is exact. Needed a new label, `ci: nightly`, which I've created.

**`permissions` is declared on the workflow** rather than inherited from the repo default (currently `write`), so tightening that default later can't silently remove the monitoring again.

### On verification

I checked one thing I'd assumed: whether the multi-line `BODY=` picks up YAML indentation and renders as a code block. It doesn't — the block scalar strips the common indent, so every line lands at column 0. Confirmed by parsing the workflow rather than by reading it.

The flake fix itself is verified by the next nightly, and **both outcomes are informative**: green proves the timeout fix, red proves the monitoring. That's the first run of this file either way.

Issue #1688, opened 2026-08-10
Carry the OSIS/USFM book code on the reference, instead of re-deriving it from the book name


bcordis:
Split out of #1623 after measuring how the providers actually identify a book. This is the real content of "mirror the standard the Bible APIs use" — and it is not about renumbering anything.

## Every provider speaks a code; none speaks a number

| provider | what goes on the wire | standard |
|---|---|---|
| API.Bible | `JHN.3.16` | **OSIS** book ID |
| Bible Brain | `/bibles/filesets/{id}/JHN/3?verse_start=16` | **USFM** book ID |
| GetBible | `Luke 7:36-38` | book *name* |
| Local (`#__bsms_bible_verses`) | `book = 43` | ours |

`ApiBibleProvider::OSIS_CODES` and `BibleBrainProvider::USFM_CODES` are **byte-for-byte identical** — 66 entries, same keys, same values, verified by parsing both arrays and comparing. OSIS and USFM agree on the canonical 66; that agreement *is* the interoperable standard, and it is a code, not a number.

Neither of our numbering schemes is a standard: **1–66** is an index into that code table (shared incidentally with GetBible's `book_nr`), and **101–173** is the same plus 100 plus the deuterocanon. See #1623 for why renumbering between them buys nothing.

## The actual defect

`ScriptureReference` carries a number and no code:

```php
public function __construct(
    public int $booknumber = 0,
    public int $chapterBegin = 0,
    …
```

So both providers throw that structure away and recover the code from the book **name**, by string matching:

```php
// ApiBibleProvider::resolveBookToOsis(), BibleBrainProvider::resolveBookToUsfm()
foreach (self::BOOK_NAMES as $num => $name) {
    if (strtolower($name) === $normalized) {
        return self::USFM_CODES[$num] ?? '';
    }
}
// then a str_starts_with() pass for abbreviations
```

A parsed reference already knows its book number. Going number → name → fuzzy match → code, to reach a table keyed by the number we started with, is a round trip through the least reliable representation available. The partial-match fallback in particular resolves "Jo" to whichever of John, Jonah, Joshua, Joel or Job it meets first.

## Proposed

1. **One code table.** Move the identical `OSIS_CODES` / `USFM_CODES` onto `AbstractBibleProvider` beside `BOOK_NAMES`, keyed by standard number. Two copies become one.
[...]

Issue #1687, opened 2026-08-10
#__bsms_books duplicates the scripture library's book table, and its published column is dead


bcordis:
Found while re-scoping #1623. Filing separately because it is independently valuable and would make #1623 smaller — bundling them would produce one unreviewable PR.

## The table holds language keys, not book names

```
booknumber  bookname
101         JBS_BBK_GENESIS
102         JBS_BBK_EXODUS
103         JBS_BBK_LEVITICUS
```

73 rows, 101–173. Every consumer JOINs it and then calls `Text::_()` on what comes back — `CwmproclaimHelper::getStudyBooks()` does exactly that:

```php
foreach ($options as $option) {
    $option->text = Text::_($option->text);
}
```

So the round trip to the database exists to fetch a translation key.

## The library already holds the same map, in code

`ScriptureHelper::BOOK_KEYS` maps the same 101–173 numbers to the same `JBS_BBK_*` keys, and `getBookName()` resolves one to a translated name. `getAllBooks()` returns the whole list, translated.

| | source | numbering | content | ever written |
|---|---|---|---|---|
| `#__bsms_books` | DB table, 73 rows | 101–173 | `JBS_BBK_*` keys | **no** |
| `ScriptureHelper::BOOK_KEYS` | library code | 101–173 | `JBS_BBK_*` keys | n/a |

Same numbering, same keys, same purpose. Nothing in the codebase inserts, updates or deletes from the table — a grep for write statements against it returns nothing.

## The published column is dead

`published TINYINT NOT NULL DEFAULT 1`, all 73 rows are `1`, **nothing filters on it**, and there is no admin view for books — no model, no controller, no table class. `BookListField`'s "published" comment refers to published *messages*, not books.

## What it costs

Roughly 20 JOINs across the admin and site code, in `CwmdescriptionHelper`, `CwmproclaimHelper`, `CwmnotificationHelper`, `BookListField`, `CwmcommentsModel`, `Cwmpodcast`, `Cwmpagebuilder`, `Cwmrelatedstudies`, `Cwmserieslist`, `Cwmlisting`, `Cwmlanding`, `CwmseriespodcastdisplayModel` and `CwmseriesdisplayModel`. Several of those are the `book2` JOINs that exist only to resolve the second flat scripture column.

There is also a correctness angle: the table cannot be translated per-site, while the library's map goes through `Text::_()` and can.

## Why this belongs before #1623 step 1

[...]

Issue #1623, opened 2026-08-06
Retire the legacy flat columns now that junction tables are the real model


bcordis:
Raised by @bcordis while reviewing #1575: *"If we need to move away from Legacy style let me know."*

Short answer: **yes, eventually — but it is a project, not a cleanup, and the interim sync-on-save pattern is the right thing until then.**

## What "legacy" means here

`#__bsms_studies` carries denormalised copies of data that now lives in junction tables:

| Flat columns on `#__bsms_studies` | Real model | Capacity |
|---|---|---|
| `teacher_id` | `#__bsms_study_teachers` | 1 teacher vs unlimited |
| `booknumber`, `chapter_begin`, `verse_begin`, `chapter_end`, `verse_end`, `bible_version` + the `…2` variants | `#__bsms_study_scriptures` | **2** references vs unlimited |

Every save has to write both and keep them agreeing. #1573 and #1575 exist precisely because those two writes were not atomic, and #1574 was the same shape on the topics junction. **That class of bug only exists because the duplication exists.**

## Measured cost of removal

Counting column *names* overstates it badly — the junction tables use the same names (`booknumber`, `chapter_begin`…), so a naive grep returns ~601 hits that are mostly legitimate junction usage.

The unambiguous signal is the **`…2` variants**, which exist only on `#__bsms_studies`:

```
107 references across 14 files
```

| refs | file |
|---|---|
| 18 | `admin/src/Lib/CwmscriptureMigration.php` |
| 16 | `admin/src/Lib/CwmpIconvert.php` |
| 12 | `admin/src/Helper/CwmscriptureHelper.php` |
| 12 | `site/src/Helper/Cwmlisting.php` |
| 10 | `site/src/Model/CwmsermonsModel.php` |
| 8 | `site/src/Helper/Cwmpagebuilder.php` |
| 6 | `admin/src/Model/CwmmessageModel.php` |
| 6 | `site/src/Model/CwmseriespodcastdisplayModel.php` |
| 6 | `site/src/Model/CwmseriesdisplayModel.php` |
| 5 | `admin/src/Table/CwmmessageTable.php` |
| 4 | `plugins/content/scripturelinks/src/Extension/ScriptureLinks.php` |
| 2 | `modules/site/mod_proclaim_youtube/src/Helper/YoutubeHelper.php` |

Note this crosses into the **scripturelinks submodule** and a **site module**, so it is not contained to the component.

## The part that argues for patience

[...]

PR #1727, opened 2026-08-11
fix: drop seven columns on #__bsms_studies that nothing reads or writes (#1690)

### LINKED ISSUE #1690: Seven columns on #__bsms_studies are declared and never used (CD/DVD production set, studytext2)
Spotted while re-scoping #1623. **Not part of that issue** — #1623 covers only the flat scripture columns and `teacher_id`, all of which are duplicated by junction tables and are still actively read and written. These are different: nothing reads or writes them at all.

## Seven columns on `#__bsms_studies` exist only as declarations

| column | definition | appears in |
|---|---|---|
| `prod_dvd` | `VARCHAR(100) DEFAULT NULL` | `CwmmessageTable`, `install.sql` |
| `prod_cd` | `VARCHAR(100) DEFAULT NULL` | same |
| `server_cd` | `VARCHAR(10) DEFAULT NULL` | same |
| `server_dvd` | `VARCHAR(10) DEFAULT NULL` | same |
| `image_cd` | `VARCHAR(10) DEFAULT NULL` | same |
| `image_dvd` | `VARCHAR(10) DEFAULT '0'` | same |
| `studytext2` | `TEXT` | same |

In `CwmmessageTable` they are bare typed properties:

```php
public ?string $prod_dvd = null;
public ?string $prod_cd = null;
public ?string $server_cd = null;
public ?string $server_dvd = null;
public ?string $image_cd = null;
public ?string $image_dvd = null;
public ?string $studytext2 = null;
```

A search across every PHP file, form XML, template, layout, JS file and language INI in the repository finds **no other reference to any of them** — no form field, no model, no query, no importer, no view. A CD/DVD production set from the era when messages shipped on physical media, plus a second study-text field, still being declared and still being carried on every row.

## They are empty on every database available

| site | studies | non-empty values across all seven |
|---|---|---|
| j5-dev | 827 | 0 |
| j6-dev | 826 | 0 |
| j6-test | 1 | 0 |
| j62-dev | 1 | 0 |
| j70-dev | 1 | 0 |

j5-dev carries real imported content, so 827 rows with nothing in any of the seven is meaningful rather than an artefact of an empty test site.

## `show_level` is a near-miss worth deciding separately

`show_level VARCHAR(100) NOT NULL DEFAULT '0'` appears in two places: `CwmmessageTable`, and `CwmpIconvert`, which **writes** it during a PreachIT import:

```php
$show_level = $pi->access;
…
$datastudies->show_level = $show_level;
```

Nothing ever reads it. So it is write-only, populated from a legacy importer, and duplicates `access` — which is the column Joomla actually uses. One non-default row exists on j5-dev and j6-dev.

Grouped here because it is the same conversation, but it needs its own call: dropping it means changing `CwmpIconvert`, which is legacy-import code that #1623 also has to leave alone until its own step 3.

## Suggested approach

1. **Confirm on a production database first.** Five development databases agreeing is good evidence, not proof — a site that ran Proclaim 7.x when DVDs were a feature may hold values. ⚠️ Same caveat #1623 carries: a third-party template override outside this repository could read these columns, and nothing here can see that.
2. Remove the seven property declarations from `CwmmessageTable`.
3. Remove them from `install.mysql.utf8.sql` — **including the seed `INSERT`** at the bottom of that file, which names all seven in its column list. Miss that and a fresh install fails on the demo study.
4. Drop the columns in a migration, one `ALTER` per column — see #1664 for why a compound `ALTER` is only ever half-checked by Joomla's schema fix, and `MigrationStatementContractTest` will fail the build if that is forgotten.
5. Decide `show_level` separately.

## Not urgent

Nothing is broken. Seven unused columns cost a little row width and a little confusion when reading the table definition. Filing it so the next person reading `CwmmessageTable` does not have to work out from scratch whether the DVD fields matter.

bcordis:
> First step of the move off the legacy Bible model. Independent of #1623, #1687 and #1688.

Seven columns on `#__bsms_studies` that exist only as declarations — the CD/DVD production set from when messages shipped on physical media, plus a second study-text field:

`prod_dvd` · `prod_cd` · `server_cd` · `server_dvd` · `image_cd` · `image_dvd` · `studytext2`

They appear in exactly three places: a property on `CwmmessageTable`, a column in `install.sql`, and a slot in the sample-data INSERT. No form field, no model, no query, no importer, no view. Empty on every database available, including j5-dev's 827 imported studies.

**Distinct from #1623**, which covers the flat scripture and teacher columns — those are duplicated by junction tables but still actively read and written. These are not read at all.

## ⚠️ One ALTER per column, deliberately

Joomla's `MysqlChangeItem` reads only words 3 and 4 of a statement, so a compound

```sql
ALTER TABLE x DROP COLUMN a, DROP COLUMN b;
```

registers as a single *"DROP COLUMN a"* — every clause after the first is neither checked by Database Maintenance nor repairable from it. Same trap as #1664. Seven statements, one per column.

## The verifier could not express a removal

`verify-migrations.php` could only assert a column *exists*. A drop is proven by absence, so it gains a `columnsAbsent` key, and 10.5.8 asserts all seven. That assertion is also what would catch a compound ALTER dropping only the first.

## Verified against real databases, and through the real updater

Applying SQL by hand proves the SQL; it says nothing about whether Joomla's migration machinery picks the file up. Both were checked.

**By hand** — the positional risk of removing seven values from a sample INSERT:

```
CREATE TABLE bsms_studies: OK      sample INSERT: OK
dead columns after install: none
row 1 landed: {"id":"1","studytitle":"Four Steps to Defeating the Flesh",
               "comments":"1","hits":"0","show_level":"0","location_id":"1"}
```

Every field after the removed block still lands in the right column — which is the failure mode of hand-editing positional values.

```
before: 58 columns, dead present: 7/7
after:  51 columns, dead present: 0/7   row intact
upgraded shape === fresh install shape: YES
```

**Through Joomla's updater** — `composer test:upgrade`, 10.5.7 → 10.5.8-dev on the role=test site:

```
PASS [10.5.8] column #__bsms_studies.prod_dvd removed
… all seven …
PASS #__schemas >= 10.5.8 (found 10.5.8-20260811)

com_proclaim  schema=10.5.8-20260811  ok=170  error=0  skipped=86
Schema OK — Joomla reports no database errors.
UPGRADE TEST PASSED 10.5.7 -> 10.5.8-dev.
```

`ok=170 error=0` is the line that matters: Joomla's own ChangeSet sees all seven drops as satisfied, which is exactly what a compound ALTER would have broken.

## Left alone on purpose

`show_level` looks similar but is not: `CwmpIconvert` **writes** it during a PreachIT import. Different question, its own change.

```
composer test:unit          1418 tests, 3018 assertions — OK
composer test:integration    442 tests, 3999 assertions — OK
composer test:upgrade        PASSED
php-cs-fixer                 0 of 632 files
```

Closes #1690


bcordis:
Checked the backup/restore path, since `view=cwmbackup` offers restore and a pre-10.5.8 backup names the dropped columns.

**Restore is safe by construction.** `Cwmrestore::restoreDB()` issues `DROP TABLE IF EXISTS` for every Proclaim table first, then replays the dump — which carries the backup's own `SHOW CREATE TABLE` output. So the INSERT column list always matches the CREATE that just ran. There is no mismatch to fail on.

**It does mean a restored old backup brings the seven columns back.** Nothing reads them, so nothing breaks — but the schema drifts from fresh-install shape, and that should be visible rather than silent. It is.

Simulated it on the role=test site by re-adding `prod_dvd` and `studytext2`:

```
com_proclaim      schema=10.5.8-20260811    ok=168  error=2   skipped=86

  ERROR in 10.5.8-20260811.sql
    statement: ALTER TABLE `#__bsms_studies` DROP COLUMN `prod_dvd`;
    check:     SHOW COLUMNS IN `#__bsms_studies` WHERE Field = 'prod_dvd'

  ERROR in 10.5.8-20260811.sql
    statement: ALTER TABLE `#__bsms_studies` DROP COLUMN `studytext2`;
    check:     SHOW COLUMNS IN `#__bsms_studies` WHERE Field = 'studytext2'
```

Joomla's own Database Maintenance detects both, names the exact statement, and offers the fix. `composer test:migrations` fails on precisely those two and passes the other five.

**This is the payoff of one ALTER per column.** A compound `ALTER ... DROP a, DROP b` would have surfaced a single error and repaired only the first, leaving the rest re-added and unreported.

Site restored to `ok=170 error=0` afterwards.

Issue #1761, opened 2026-08-13
com_proclaim ships 511 duplicate JBS_BBK_* book-name keys that nothing reads


bcordis:
## Summary

`com_proclaim` ships **73 `JBS_BBK_*` book-name keys in each of seven language files** — 511 in total — that nothing reads. `lib_cwmscripture` owns these strings and loads them itself.

Follow-up to the comments on #1688.

## Where they are

`admin/language/<loc>/<loc>.com_proclaim.ini` for cs-CZ, de-DE, en-GB, es-ES, hu-HU, nl-NL, no-NO. (There is no `site/language`, and the other seven locales have never carried book names in either extension.)

## Why they are dead

The library owns the whole book-name domain — `BOOK_KEYS`, the 187-entry `ABBREVIATIONS` table, and `getBookName()` / `getBookNumber()` / `getAbbreviations()` — and, importantly, **loads its own language file**:

```php
Factory::getApplication()->getLanguage()
    ->load('lib_cwmscripture', JPATH_LIBRARIES . '/cwmscripture');
```

That removed the old dependency where the keys resolved only because com_proclaim's admin language happened to be loaded.

On the Proclaim side the migration is already complete:

- no PHP under `admin/src`, `site/src` or `api/src` references `JBS_BBK_*` — the only matches are docblocks in `Cwmlisting.php:2181` and `Cwmlanding.php:1194` describing the old behaviour
- eight files delegate to `ScriptureHelper`
- `#__bsms_books` was dropped in 10.5.8 (#1687); the two `type="SQL"` form fields became `BookListField`
- nothing outside the language files references the keys (the only other hit is a comment in `admin/sql/updates/mysql/10.5.8-20260811.sql`)

I diffed all seven locales against the library's copies: **values are identical**, differing only in formatting (`KEY = "v"` here, `KEY="v"` in the library).

## Why remove rather than leave

1. **Silent precedence.** Both extensions define the same keys into one global `Language` object. If they ever diverge, whichever loads last wins, and nothing surfaces which one rendered.
2. **It misleads translators.** Someone editing book names in `com_proclaim.ini` would reasonably believe they are editing the strings users see. They are not — the library's copies are.

## ⚠️ Decide the fallback behaviour first

There is **no version gate**. `LibraryVersion::satisfies()` and `needsUpgrade()` have no callers anywhere — only README examples and the library's own test. Proclaim relies on its package installing and locking the library instead.

[...]

PR #1766, opened 2026-08-13
fix: ask the library for the book name instead of translating the key

bcordis:
Groundwork for #1761. **Does not close it** — see below.

## The change

`formReference()` called `Text::_($row->bookname)` directly on a `JBS_BBK_*` key. That only resolves if something already loaded a language file carrying those keys — and the thing satisfying it was **com_proclaim's own duplicate copy**, which #1761 proposes deleting.

`ScriptureHelper::getBookName()` loads `lib_cwmscripture`'s language file itself, so asking the library removes the hidden dependency on load order. Falls back to the old call when the row has no book number, since legacy rows predate it.

## Why #1761 is still open

Deleting the 511 duplicate keys fails four characterisation tests:

```
-'John 3'
+'JBS_BBK_JOHN 3'
```

`ScriptureHelper::loadLanguage()` reads from `JPATH_LIBRARIES . '/cwmscripture'` — the **installed** path — while the repo keeps the library at `libraries/lib_cwmscripture`. Under test the load fails silently and `Text::_()` returns the key.

So the suite **cannot currently prove the removal is safe**, and the duplicates are what has been making those tests pass. On a real install the path exists and the library's own copies would resolve, but "probably fine in production" is not something to ship on when the tests say otherwise.

The removal needs the test environment able to load the library's language first. That is the actual blocker on #1761, and it was not visible until the keys came out.

## Note on the original survey

I previously checked for remaining `JBS_BBK_` references and found only docblocks. That grep could not see this one: the key arrives in `$row->bookname` as a variable, so the string never appears at the call site. Worth remembering for the next "nothing references this" claim.

## Testing

2285 tests, 7606 assertions, all passing. PSR-12 clean across 646 files.


bcordis:
## Correction: the path explanation in the description is not established

The description says the removal fails because `ScriptureHelper::loadLanguage()` reads `JPATH_LIBRARIES . '/cwmscripture'` while the repo keeps the library at `libraries/lib_cwmscripture`. **I could not confirm that, and the evidence points away from it.**

What actually checks out:

- `tests/unit/bootstrap.php:166` sets `JPATH_LIBRARIES` to `<joomla-cms-path>/libraries` — a real Joomla install, not the repo.
- That install *does* have the library: `Sites/j5-dev/libraries/cwmscripture -> ../../../GitHub/lib_cwmscripture`.
- That symlink target *does* carry the keys — 73 `JBS_BBK_*` entries in its en-GB file.

So the path resolves and the file exists. The load failing is more likely about the language/application setup under test — `loadLanguage()` wraps its call in `try { Factory::getApplication()->getLanguage()->load(...) } catch (\Throwable) {}`, so with no application it swallows the failure and `Text::_()` returns the key. The bootstrap does no explicit language loading at all, which fits.

**Do not start the bootstrap fix from the path theory in the description.** Verify where the load actually fails first — most directly by calling `ScriptureHelper::getBookName(143)` inside the failing test and seeing whether `loadLanguage()` reaches its `catch`.

⚠️ Separately worth knowing: that symlink points at the **standalone clone** `GitHub/lib_cwmscripture`, not the submodule inside Proclaim. The standalone clones are stale — so the dev site, and therefore the test run, may not be exercising the library code merged today.

The change in this PR is unaffected by any of it: asking the library for the name is correct regardless, and the suite is green.

bcordis:
## Second commit: the removal itself — this now closes #1761

The earlier note said the path theory was unconfirmed. **It was right after all** — I had verified against the wrong install. Running the check inside the failing test settled it:

```
JPATH_LIBRARIES = .../GitHub/joomla-cms/libraries
lib dir exists  = NO
load() returned = false
Text::_(JBS_BBK_JOHN) = JBS_BBK_JOHN
```

Tests run against `GitHub/joomla-cms`, a bare checkout with no library installed. My earlier check looked at `j5-dev`, which *does* carry the symlink — hence the false correction.

`bootstrap.php` now creates that link, and the same check returns `John`.

## What is in this commit

- **511 keys removed** from the seven `com_proclaim.ini` files that carried them
- **Bootstrap** links `libraries/lib_cwmscripture` → `JPATH_LIBRARIES/cwmscripture`
- **Four assertions corrected** in `CwmscriptureHelperTest`

That last one is the part worth reviewing. They asserted `'JBS_BBK_LUKE 7:36-38'`, with a comment explaining that "in test context, Text::_() returns the language key as-is" — framing a missing library as a property of testing. They would have passed forever while asserting that scripture headings render as raw keys.

## Verified before removing

Content comparison, not counts — every key present in the library with an identical value:

```
cs-CZ  proclaim=73 lib=73 missing=0 value-diffs=0
de-DE  proclaim=73 lib=73 missing=0 value-diffs=0
en-GB  proclaim=73 lib=73 missing=0 value-diffs=0
es-ES  proclaim=73 lib=73 missing=0 value-diffs=0
hu-HU  proclaim=73 lib=73 missing=0 value-diffs=0
nl-NL  proclaim=73 lib=73 missing=0 value-diffs=0
no-NO  proclaim=73 lib=73 missing=0 value-diffs=0

TOTAL KEYS LOST: 0
```

The seven locales the library ships without book names (fr-FR, it-IT, ja-JP, ko-KR, pt-BR, ru-RU, zh-CN) never had them here either — they fall back to en-GB through Joomla's language merge, unchanged.

## Behaviour change, deliberate

A site whose scripture library is missing or failed to install now shows raw keys instead of book names. Confirmed as intended: a broken install should look broken, and every other scripture feature is already gone in that state. The duplicates were masking it.

2285 tests, 7606 assertions, green. PSR-12 clean across 646 files.

Issue #28, opened 2012-10-18
Migration: Access level of series is not set correctly


st-anton:
During migration, the series get the access = 0, but this is not a valid value. Result is, that a study is not displayed, if it is part of a serie. (see models/sermon.php line 189). This may or may not happen because I defined a new access leven (like members of our church and friends of our church), so I have more access levels than the predefined levels.
Suggested solution:
- Define the default access level to public or registered, like as it is done during migration of studies.
- Additionally, the admin must be given a hint / forum entry, that something like this may happen, if more than the standard access levels are defined.

PR #1770, opened 2026-08-13
chore: refresh cwm/build-tools to 1.18.0 for the artifact token gate

bcordis:
Picks up `cwm/build-tools` **1.17.0** (cwm-build-tools#85 — the release gate that unzips the artifact before pushing and refuses to publish if `__DEPLOY_VERSION__` survived, recursing into nested zips). The lock was pinned at **1.15.1** under a `^1.12.0` floor; both move to `^1.17.0`.

## ⚠️ Do not merge before cwm-build-tools#89

Proclaim bundles `pkg_cwmscripture` as a **`subBuild` of a git submodule**, and `subBuild` packages the child from its working tree *without substituting* — substitution is a release step in the child's own repo, which a `subBuild` never runs. The outer release must not substitute into a submodule either (it would stamp another repo's source with Proclaim's version, and since 1.16.0 the substituter refuses).

So the bundled child contributes occurrences that **no Proclaim-side change can remove**. From a locally built `pkg_proclaim-10.5.8-dev.zip`:

```
10  ...!packages/pkg_cwmscripture.zip!script.install.php
 5  ...!packages/pkg_cwmscripture.zip!plg_system_cwmscripture.zip!src/Extension/Cwmscripture.php
 1  ...!packages/pkg_cwmscripture.zip!plg_content_scripturelinks.zip!script.php
```

Merging this arms the gate and **stops Proclaim's next release** with no remedy available in this repo. That is why it is held.

To be clear, the gate is not wrong: the shipped `pkg_proclaim-10.5.7.zip` really does carry **23** literal tags, 16 of them from that path. The problem is that Proclaim currently has no way to fix them.

## Verified

- Resolves to v1.17.0; the tree lands in `libraries/vendor/` (this repo sets `vendor-dir`) and carries the scanner, the CLI and the `release.sh` wiring.
- Ran the shipped gate against real artifacts: `pkg_proclaim-10.5.7.zip` → 23 findings across three nesting levels; `pkg_proclaim-10.5.8-dev.zip` → the 16 above plus `com_proclaim`'s own placeholders, which a real release substitutes (`admin/`, `site/`, `api/` are all in `paths`) and which are expected in a dev build.

## Recommended order

1. cwm-build-tools#89 — substitute a `subBuild` child using the child's own version before packaging
2. release that as 1.18.0
3. re-point this PR at it, then merge

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,177 +1,239 @@
 {
-  "name": "joomla-bible-study/proclaim",
-  "type": "project",
-  "description": "CWM Proclaim",
-  "keywords": [
-    "bible",
-    "joomla",
-    "comments",
-    "Proclaim"
-  ],
-  "homepage": "https://github.com/Joomla-Bible-Study/Proclaim",
-  "license": "GPL-2.0-or-later",
-  "authors": [
-    {
-      "name": "Tom Fuller",
-      "email": "info@christianwebministries.org",
-      "homepage": "https://www.christianwebministries.org",
-      "role": "Tech Writer & Developer"
-    },
-    {
-      "name": "Brent Cordis",
-      "email": "info@christianwebministries.org",
-      "homepage": "https://www.christianwebministries.org",
-      "role": "Lead Developer"
-    }
-  ],
-  "minimum-stability": "dev",
-  "prefer-stable": true,
-  "repositories": [
-    {
-      "type": "vcs",
-      "url": "https://github.com/Joomla-Bible-Study/cwm-build-tools"
-    }
-  ],
-  "config": {
-    "optimize-autoloader": true,
-    "platform": {
-      "php": "8.3.0"
-    },
-    "vendor-dir": "libraries/vendor",
-    "github-protocols": [
-      "https"
+    "name": "joomla-bible-study/proclaim",
+    "type": "project",
+    "description": "CWM Proclaim",
+    "keywords": [
+        "bible",
+        "joomla",
+        "comments",
+        "Proclaim"
     ],
-    "allow-plugins": {}
-  },
-  "support": {
-    "issues": "https://www.christianwebministries.org/issues",
-    "forum": "https://www.christianwebministries.org/forum",
-    "docs": "https://www.christianwebministries.org/docs"
-  },
-  "autoload": {
-    "psr-4": {
-      "CWM\\Component\\Proclaim\\Administrator\\": "admin/src/",
-      "CWM\\Component\\Proclaim\\Api\\": "api/src/",
-      "CWM\\Component\\Proclaim\\Site\\": "site/src/",
-      "CWM\\Library\\Scripture\\": "libraries/lib_cwmscripture/src/"
-    },
-    "exclude-from-classmap": [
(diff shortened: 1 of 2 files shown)


bcordis:
## Correction: 10 unfixable, not 16 — and the blocker is now fixed upstream

Two updates.

**The count in the description was overstated.** Of the 16 occurrences the child contributes, **6 are already fixed in the child's source** — CWMScriptureLinks#28 set `plg_system_cwmscripture/src/Extension/Cwmscripture.php` (5) and root `script.php` (1) to `@since 1.2.0`. They persist here only because this repo pins `v1.2.5`, which predates that PR:

```
v1.2.5      plg_system + script.php → 6      build/script.install.php → 10
origin/main plg_system + script.php → 0      build/script.install.php → 10
```

So a re-pin to a post-#28 child tag removes 6 with no tooling change, and only the **10 installer tokens** were structural. "No Proclaim-side change can remove" was true of 10, not 16.

**The structural 10 are fixed upstream.** cwm-build-tools#89 is closed by PR #91 (`e839bc4`): a `subBuild` child is now substituted with **its own** version before packaging and restored byte-for-byte afterwards. Verified against this exact child — version resolves to `1.2.5`, the ten installer tokens become `@since 1.2.5`, restore puts all ten back.

**Still held**, but only on the release: #91 is on `main` and not yet in a tagged build-tools version. Order is unchanged — cut 1.18.0, re-point this PR at it, then merge.

bcordis:
## Hold lifted — re-pointed at 1.18.0

cwm-build-tools **1.18.0** is released and this PR now moves the floor to `^1.18.0` (was `^1.17.0`).

1.17.0 alone armed the gate without giving Proclaim a way past it. 1.18.0 adds the other half: a `subBuild` child is substituted with **its own** version before packaging and its tree restored byte-for-byte afterwards (cwm-build-tools#89, PR #91).

### What was verified

Running the shipped 1.18.0 code against this repo's actual child (`plugins/content/scripturelinks` at the pinned `v1.2.5`):

```
child version resolved from its manifest   1.2.5
build/script.install.php                   10 replacement(s) → @since 1.2.5
after restore                              10 tokens back, tree unchanged
```

That is exactly the structural set that previously had no fix.

### What was NOT verified

**A full local `pkg_proclaim` build was not completed.** The child's own nested library build needs npm-built assets that my scratch clone did not have, so I could not produce and scan an end-to-end `pkg_proclaim-*.zip`. The substitution and restore behaviour — the part 1.18.0 changes — was verified directly, and it is covered by 11 tests upstream including a full `subBuild` that packages a child and asserts it carries the child's version and not the wrapper's. But the first real `composer release` here will be the first full-stack exercise of the gate.

### Still worth doing separately

The pin is `v1.2.5`, which predates CWMScriptureLinks#28 — a re-pin to a newer child tag removes 6 further occurrences (`plg_system_cwmscripture` ×5, root `script.php` ×1) that are already fixed in the child's source. Not required for this PR, and not required for the gate to pass now, but it is free cleanliness whenever the scripture stack is next re-pinned.

Issue #93, opened 2013-02-13
Uninstall does not remove tables


tomfuller2:
For some reason in biblestudy.script.php the uninstall function - the db query returned the manifest of the component from the extensions table. 

So I changed the query to be completely static. That fixed it. No worries of SQL injection as there are no variables in the query.

Now the tables uninstall correctly.

Brent - you had also had "parent_id" spelled as "parent_id" in one query so I fixed that too.

https://github.com/Joomla-Bible-Study/Joomla-Bible-Study/commit/36e0f34b1957ee8be466c4963b945df1e6a2731c

PR #1771, opened 2026-08-13
chore: restore composer.json formatting, and move to build-tools 1.18.1

bcordis:
#1770 changed one constraint but rewrote all 402 lines of `composer.json`: I edited it by parsing the JSON and dumping it back, which normalised the file's **2-space indentation to 4-space**.

The content was never altered — parsing both revisions and comparing gives an exact match apart from the constraint — but 232 lines of formatting churn buried the real change and would poison `git blame` on every line of the file.

This restores the file byte-for-byte from `a16e55cfc` with the single constraint line edited in place:

```diff
-    "cwm/build-tools": "^1.12.0",
+    "cwm/build-tools": "^1.18.0",
```

`composer.lock` is untouched — its `content-hash` is computed from the dependency fields, not the file's formatting — and `composer validate` passes.

CWMScriptureLinks was unaffected (already 4-space); lib_cwmscripture has the same 83-line churn and is being fixed in its own PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,239 +1,177 @@
 {
-    "name": "joomla-bible-study/proclaim",
-    "type": "project",
-    "description": "CWM Proclaim",
-    "keywords": [
-        "bible",
-        "joomla",
-        "comments",
-        "Proclaim"
-    ],
-    "homepage": "https://github.com/Joomla-Bible-Study/Proclaim",
-    "license": "GPL-2.0-or-later",
-    "authors": [
-        {
-            "name": "Tom Fuller",
-            "email": "info@christianwebministries.org",
-            "homepage": "https://www.christianwebministries.org",
-            "role": "Tech Writer & Developer"
-        },
-        {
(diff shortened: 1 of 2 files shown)


bcordis:
## Also bumps to 1.18.1 — `development` is currently broken for packaging

1.18.0 shipped `ChildTokenSubstitution` wired into the packager but **not required** by `scripts/package.php`, which has no autoloader. `cwm-package` therefore fatals with `Class ... not found` on any project with a `subBuild` include — and Proclaim has one (`pkg_cwmscripture`).

#1770 locked `development` at exactly that version, so **packaging is broken on `development` right now**. Fixed upstream in cwm-build-tools 1.18.1 (PR #93); this PR moves the floor to `^1.18.1` and refreshes the lock.

The formatting restore is unchanged — still exactly one line different from the pre-churn `a16e55cfc`:

```diff
-    "cwm/build-tools": "^1.12.0",
+    "cwm/build-tools": "^1.18.1",
```

PR #1794, opened 2026-08-14
fix: let Batch → Teacher add rather than replace

### LINKED ISSUE #1793: Batch → Set Teacher silently drops a message's other teachers
Messages have had multiple teachers since the `#__bsms_study_teachers` junction landed, but the batch action still treats the teacher as a scalar.

## What happens

`CwmmessageModel::batchTeacher()` passes the chosen teacher as if it were the complete set:

```php
$teachers = $teacherId > 0 ? [['teacher_id' => $teacherId]] : [];
CwmstudyteacherHelper::saveTeachers((int) $pk, $teachers);
```

`saveTeachers()` → `writeTeachers()` deletes every junction row for that study and re-inserts what it was handed:

```php
$query = $db->createQuery()
    ->delete($db->quoteName('#__bsms_study_teachers'))
    ->where($db->quoteName('study_id') . ' = ' . $studyId);
```

That is **correct for the edit form**, where the subform genuinely is the full set. It is wrong for batch: selecting a teacher for a message that already credits two replaces both with one. No warning, no undo, and the batch reports "Batch process completed".

## How core does it

`AdminModel::batchTags()` dispatches with `'replaceTags' => false` — it **adds** to what is there — and carries a separate `$removeTags` flag for the opposite intent.

Proclaim's teacher batch offers neither: it cannot add, and it cannot clear (unlike Location, which does have `- Clear Location -`). Replace-with-exactly-one is the only thing it can do, and that is not what `- Do Not Change Teacher -` leads you to expect.

## Blast radius

Real but not widespread. On a ~836-message dataset, 3 studies have more than one teacher (max 2). The concern is that the loss is **silent and unrecoverable** — nothing records what was there before, so it surfaces only when someone notices a missing credit later.

## Options

1. **Add mode** — batch appends the teacher, matching core's tag behaviour. Safest default, but then there is no way to *set* a single teacher.
2. **Two actions** — "Add teacher" and "Replace teachers", the shape core arrives at with add/remove.
3. **Multi-select** — make the batch field `multiple`, so what you pick honestly *is* the full replacement set. Keeps one action and makes the replace semantic visible instead of hidden.

3 is the smallest change that stops the surprise; 2 is the most flexible. Worth deciding before anyone leans on batch for teacher cleanup.

## Not a regression

Pre-existing. The junction write predates the `teacher_id` retirement work in #1791, which only removed a dead flat-column assignment beside it and left these semantics untouched.

bcordis:
Closes #1793.

A message can credit several teachers, but the batch handed its single selection straight to `saveTeachers()`, which rewrites the whole junction. Setting a teacher on a message that already credited two silently dropped the other, and the batch reported "Batch process completed".

## What changed

The batch now carries a mode, read from the commands the way core's `batchTag` reads `tag_addremove`, and **defaults to adding**:

```
Teacher
┌────────────────────────────────────┐
│ - Do Not Change Teacher -       ▾  │
└────────────────────────────────────┘
  (•) Add to existing teachers
  ( ) Replace all teachers
```

- **Add** (default) — merges into the existing credits, and is idempotent if the teacher is already there.
- **Replace** — the old behaviour, still available, but now chosen rather than assumed.

`resolveTeacherMode()` treats anything other than a literal `'r'` as add, so a form post or script that omits the command cannot reach the destructive branch.

## Replace now says what it destroyed

Per Brent's note on the issue — when Replace actually discards someone, the batch warns instead of reporting a clean success:

> ⚠️ *3 messages credited other teachers. Replacing removed them.*

The count comes from `getTeachersForStudies()`, which is a single query for the whole selection, so this costs nothing per row.

## Why a mode and not multi-select

All three options on the issue were buildable — `batch[]` is read as an array, so a `multiple` select would have worked. The mode was chosen because it is the only one where the **default is non-destructive**: with multi-select the semantics are still replace, just more visible. Add-by-default means the accident is no longer reachable without a deliberate click.

## Tests

`CwmbatchTeacherTest` runs against the real junction: add keeps existing credits, add is idempotent, replace still replaces, and only a literal `'r'` selects it.

**Both halves mutation-verified** — pinning `$replacing = true` (the old behaviour) fails the two add tests; defaulting the mode to `'r'` fails the resolver test.

⚠️ The tests delete their rows by hand rather than rolling back: `batchTeacher()` calls `Table::store()`, and `Table::store()` commits the transaction.

`composer test` — 2349 green; lint clean.

## Not covered here

Batch still cannot **clear** a message's teachers — an empty command is skipped by `AdminModel::batch()` before any of this runs, exactly as before. Same for Location, which papers over it with a `- Clear Location -` option. Out of scope; say the word and it is a small follow-up.

### DIFF
--- a/admin/language/en-GB/en-GB.com_proclaim.ini
+++ b/admin/language/en-GB/en-GB.com_proclaim.ini
@@ -1199,7 +1199,12 @@ JBS_BAT_LOCATION_NOCHANGE = "- Do Not Change Location -"
 JBS_BAT_SERIES_DESC = "Choose a new series"
 JBS_BAT_SERIES_NOCHANGE = "- Do Not Change Series -"
 JBS_BAT_TEACHER_DESC = "Choose a teacher for the batch process"
+JBS_BAT_TEACHER_MODE_ADD = "Add to existing teachers"
+JBS_BAT_TEACHER_MODE_QUESTION = "How should this teacher be applied?"
+JBS_BAT_TEACHER_MODE_REPLACE = "Replace all teachers"
 JBS_BAT_TEACHER_NOCHANGE = "- Do Not Change Teacher -"
+JBS_BAT_TEACHER_REPLACED_N_1 = "%d message credited other teachers. Replacing removed them."
+JBS_BAT_TEACHER_REPLACED_N_MORE = "%d messages credited other teachers. Replacing removed them."
 
 
 ;; BBK Book of the Bibles Translations
--- a/admin/src/Helper/Cwmhtml.php
+++ b/admin/src/Helper/Cwmhtml.php
@@ -217,6 +217,21 @@ class Cwmhtml
             '<option value="">' . Text::_('JBS_BAT_TEACHER_NOCHANGE') . '</option>',
             HTMLHelper::_('select.options', self::teacherList(), 'value', 'text'),
             '</select>',
+            '<fieldset id="batch-teacher-mode" class="mt-2">',
+            '<legend class="visually-hidden">' . Text::_('JBS_BAT_TEACHER_MODE_QUESTION') . '</legend>',
+            HTMLHelper::_(
+                'select.radiolist',
+                [
+                    HTMLHelper::_('select.option', 'a', Text::_('JBS_BAT_TEACHER_MODE_ADD')),
+                    HTMLHelper::_('select.option', 'r', Text::_('JBS_BAT_TEACHER_MODE_REPLACE')),
+                ],
+                'batch[teacher_mode]',
+                '',
+                'value',
+                'text',
+                'a'
+            ),
+            '</fieldset>',
         ];
 
         return implode("\n", $lines);
--- a/admin/src/Model/CwmmessageModel.php
+++ b/admin/src/Model/CwmmessageModel.php
@@ -80,6 +80,56 @@ class CwmmessageModel extends AdminModel
     ];
 
     /**
+     * Whether a batched teacher is added to the message's existing teachers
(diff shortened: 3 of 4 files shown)


bcordis:
## Live verification on j6-dev

Driven through the real batch dialog, not the model — every step confirmed against the junction table afterwards.

**Widget renders as designed.** Legend "How should this teacher be applied?", both radios present, and **Add is checked on load** — so the shipped default is the non-destructive one.

| step | before | action | after | result |
|---|---|---|---|---|
| Add keeps existing | `[2]` | add teacher 3, mode left at its default | `[2, 3]` | ✅ the bug, fixed |
| Add is idempotent | `[2, 3]` | add teacher 3 again | `[2, 3]` | ✅ no duplicate row, no reordering |
| Replace still replaces | `[2, 3]` | replace with teacher 2 | `[2]` | ✅ |

**Both plural forms exercised**, since a wrong `_MORE` key would only show up at scale:

- one message → `⚠️ 1 message credited other teachers. Replacing removed them.`
- two messages → `⚠️ 2 messages credited other teachers. Replacing removed them.`

Add reports a plain `Success — Batch process completed.` with **no** warning, which is right: nothing was discarded.

All test rows restored afterwards; the dataset is back to its starting counts (836 studies with a teacher, 3 with more than one).

<details>
<summary>One thing worth knowing if you script against this list</summary>

Selecting a row by checkbox value alone picks the wrong element — the messages list also renders `table[column][]` inputs that collide on value. `input[type=checkbox][name="cid[]"][value="N"]` is the reliable selector. Cost me one run that failed with "No items selected".
</details>

PR #1797, opened 2026-08-14
fix: derive comment moderation state server-side

### LINKED ISSUE #1796: Comment submission: a visitor can approve their own comment, and the captcha never runs
`CwmsermonModel::storecomment()` takes the moderation state straight from the request:

```php
'published' => $input->getInt('published', 1),
```

The comment form posts it as a hidden field seeded from the template's `comment_publish`, but a visitor decides what a visitor posts. Nothing re-derives it server-side, and `CwmcommentTable::check()` only fills `user_id` — it never touches `published`.

## Impact

**Anyone can approve their own comment on a site configured to hold them.** Two details make it worse:

- **The default is `1`.** Omitting the field entirely publishes.
- **`comment()` performs no access check.** `comment_access` gates whether the form *renders*, not whether the handler accepts a POST.

The shipped default template has `"comment_publish":"0"`, so moderation is on by default — and bypassable by default. Spam posts straight to live pages.

### Verified, not inferred

On a site with `comment_publish=0`, a crafted POST carrying `published=1` stored a **visible** comment:

```
comment id=3  published=1     <- site setting says hold for approval
```

## Two more defects, same area

**1. The captcha never ran.** `CwmsermonController::comment()` reads template params like this:

```php
if (!empty($model->_template[0]->params)) {
    $params->loadString($model->_template[0]->params);
}
```

`_template` is **never assigned anywhere in the codebase** — it appears only on those two lines. Behind `!empty()` that is silent, so `$params` was always empty and every setting took its default:

- `use_captcha` → unset → **the captcha block never executed**, however the template was configured
- `email_comments` → unset → **no notification email was ever sent**
- `comment_notify_email` → never read

**2. A mail failure 500s the visitor.** Once params load correctly the notification path runs for the first time, and `$mail->Send()` throwing takes the whole request with it — after the comment has already been stored. The visitor sees an error for a comment that saved, and resubmits. (I produced four duplicate rows this way while testing.)

## Fix

- Derive `published` from the template's `comment_publish`, never from the request, defaulting to **0** to match `template.xml`'s own `default="0"` — an unreadable setting should hold, not publish.
- Set `comment_date` server-side.
- Load the template params through `Cwmparams::getTemplateparams()` so the captcha and notifications actually work.
- Wrap notification sends so a mail failure cannot fail a stored comment.
- Drop the `published` / `comment_date` hidden fields — posting them only invited forgery.

## Test coverage before this

The whole Comments system had **3 test methods**: one that the batch resolves a model name, and two that `check()` handles `user_id`. Nothing covered public submission.

bcordis:
Closes #1796.

A visitor could approve their own comment. `storecomment()` read the moderation state from the request, defaulting to **published**, and nothing re-derived it server-side.

**Verified before the fix**, on a site set to hold comments — a POST carrying `published=1` stored a visible comment. The same request now stores it held.

## What changed

| | |
|---|---|
| `published` | from the template's `comment_publish`, default **0** (matching `template.xml`) |
| `comment_date` | server-side |
| template params | loaded via `Cwmparams::getTemplateparams()` |
| notifications | can no longer fail a stored comment |
| hidden `published`/`comment_date` fields | removed |

## The captcha never ran

`comment()` read `$model->_template[0]->params` — a property **nothing in the codebase ever assigns**, appearing only on those two lines. Behind `!empty()` that is silent, so `$params` was always empty and every setting took its default: no captcha check however the template was configured, and no notification email ever sent.

Fixing that made the notification path run for the first time, which exposed the third defect — a throwing mail transport took the whole request with it *after* the comment was stored, so the visitor saw an error for a comment that had saved, and resubmitted. I generated four duplicate rows that way while testing.

## Tests

The Comments system had **3 test methods** before this, none covering public submission. Five integration tests now exercise it against the real table, including the exploit itself, plus fail-closed on an unset setting and the forged-date case.

**Mutation-verified:** restoring the request read fails four of the five.

`composer test` 2354 green; lint clean.

### DIFF
--- a/site/src/Controller/CwmsermonController.php
+++ b/site/src/Controller/CwmsermonController.php
@@ -17,6 +17,7 @@ use CWM\Component\Proclaim\Site\Model\CwmsermonModel;
 use Joomla\CMS\Captcha\Captcha;
 use Joomla\CMS\Factory;
 use Joomla\CMS\Language\Text;
+use Joomla\CMS\Log\Log;
 use Joomla\CMS\MVC\Controller\FormController;
 use Joomla\CMS\MVC\Model\BaseDatabaseModel;
 use Joomla\CMS\Router\Route;
@@ -174,12 +175,11 @@ class CwmsermonController extends FormController
         $model = $this->getModel();
         $t     = $input->getInt('t', 1);
 
-        // Get template params
-        $params = new Registry();
-
-        if (!empty($model->_template[0]->params)) {
-            $params->loadString($model->_template[0]->params);
-        }
+        // ⚠️ This read $model->_template[0]->params, a property nothing ever
+        // sets. Behind !empty() that is silent, so $params was always empty and
+        // every setting below took its default: the captcha was never checked
+        // however the template was configured, and no notification was sent.
+        $params = $model->getCommentParams();
 
         // Check captcha if enabled (Joomla 5/6 compatible Captcha API)
         if ($params->get('use_captcha') > 0) {
@@ -207,7 +207,9 @@ class CwmsermonController extends FormController
 
         // Store the comment
         if ($model->storecomment()) {
-            $published = $input->getInt('published', 1);
+            // The state the comment was actually stored with, not the one the
+            // request asked for.
+            $published = $model->getCommentPublishState();
 
             // Show appropriate message based on whether the comment is auto-approved or held
             if ($published === 0) {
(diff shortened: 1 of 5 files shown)


bcordis:
Added a second commit while testing this on j6-dev.

**Show/Hide Comments did nothing.** The comments block renders inside a Bootstrap collapse, but `Cwmsermon` never loaded `bootstrap.collapse` — only the landing page view does. Since the section starts with `class="collapse"` (hidden), the comments could not be opened **at all**, not just that the animation was missing.

Verified on a sermon page: the button now toggles `#JBScomments` from `collapse` to `collapse show` and the section becomes visible.

Separately, and **not** a bug: "cannot comment" on a site with `comment_access=2` is the Registered gate working as configured — a guest correctly gets the `JBS_CMT_REGISTER_TO_POST_COMMENTS` lock message instead of the form.

PR #1861, opened 2026-08-16
fix: restore the scripture stack after clearing consumers, instead of guarding it

### LINKED ISSUE #1860: test:upgrade phase 10 destroys the scripture stack, and the damage outlives the run
`composer test:upgrade` phase 10 clears "other scripture consumers" before the destructive phases. On a site carrying one, that removal takes the shared scripture stack with it — and because the removal persists, every later run starts from the wreckage.

This is the failure #1820 was meant to close. It is not closed: **`package_id` is no longer the vector.**

## Observed

Full run against 10.5.9 → 10.5.10-dev, 2026-08-16:

```
-- [10/17] verify the consumer registry schema landed and the shipped extensions registered
  OK   com_proclaim registered itself as a consumer
  OK   scripturelinks (content) registered itself as a consumer
  OK   cwmscripture (task) registered itself as a consumer
   protecting the scripture stack from the removals below:
  scripture stack already owned by pkg_cwmscripture      <-- 0 rows needed reparenting
   clearing other scripture consumers before the uninstall phases:
     removed extension id 4274
-- [11/17] STILL NEEDED: removing the library alone must be refused
No library 'cwmscripture' registered in #__extensions.
```

Afterwards, three rows were gone from `#__extensions`:

| extension | row | files on disk |
|---|---|---|
| `lib_cwmscripture` (library `cwmscripture`) | **gone** | **still present** |
| `plg_content_scripturelinks` | **gone** | gone |
| `plg_task_cwmscripture` | **gone** | gone |

`plg_system_cwmscripture` and `pkg_cwmscripture` survived.

## Why the existing guard does not cover it

`verify-scripture-uninstall.php`'s `protect-scripture-ownership` re-points `package_id` at `pkg_cwmscripture` for the library and the content, task and system plugins, then reports how many it moved. It printed **"scripture stack already owned by pkg_cwmscripture"** — `$moved === 0`, meaning all four rows were *already* correctly parented before anything was removed.

So the removal did not reach them as Joomla package children. Re-parenting is necessary but not sufficient.

The likely remaining vector is the other package's **own uninstall script** removing them by element, the way `pkg_proclaim`'s `script.install.php::uninstall()` does for exactly these extensions. Not confirmed — the LivingWord rows were destroyed by the run, so there was nothing left to read.

## Why it matters beyond one red run

- **It silently disarms phases 11–17.** Every one of them needs a library to test against, so a defect in the uninstall guards would now surface as "no library registered" rather than as itself.
- **It persists.** `reset-testsite.php` only clears the Proclaim family, so the next run fails at phase 11 the same way. `test-upgrade.sh` says this outright in the phase-10 comment; the safeguard it describes is the one that just failed to hold.
- **It quietly removes the fixture.** The cleared consumer stays cleared, so the destructive path is *disarmed for subsequent runs* — a green run afterwards proves less than it appears to. That is exactly what happened here: after reinstalling `pkg_cwmscripture` the re-run passed all 17 phases, but only because `other-consumer-ids` returned empty and phase 10 never entered the branch.

## Diagnostic tell

⚠️ `libraries/cwmscripture/` **survives on disk with no `#__extensions` row** — an orphaned library directory. The two plugin directories go with their rows. Check disk and DB, not just DB.

## Repair

```bash
php <site>/cli/joomla.php extension:install \
  --path=plugins/content/scripturelinks/build/dist/pkg_cwmscripture-<version>.zip
```

Restores the library and both plugins, correctly parented to `pkg_cwmscripture`. LivingWord, once cleared, needs reinstalling by hand.

## Suggested direction

1. Reproduce with a LivingWord fixture present and capture *what* removes the three rows before it is destroyed — the uninstall script is the first place to look.
2. Make phase 10 verify the stack is still installed **after** clearing consumers, and fail loudly there. Failing at the point of damage beats failing one phase later with a misleading message.
3. Consider snapshotting and restoring the stack around the clearing step, rather than relying on `package_id` alone to protect it.

bcordis:
Fixes #1860.

## The guard could never have worked

Phase 10 of `test:upgrade` clears other scripture consumers before the destructive phases, and that removal took `lib_cwmscripture`, `plg_content_scripturelinks` and `plg_task_cwmscripture` with it. Phase 11 then died on `No library 'cwmscripture' registered`, and because `reset-testsite.php` only clears the Proclaim family, the next run started from the same wreckage.

`protect-scripture-ownership` is **removed rather than extended**. `PackageAdapter::removeExtensionFiles()` reads the consumer's *installed* manifest from `administrator/manifests/packages` and resolves each declared `<file>` through `_getExtensionId($type, $element, $client, $group)` — **by element**. `package_id` is never consulted on that path, so re-pointing it corrects a column Joomla does not read here.

`build/pkg_livingword.xml` declares all three as its own children and has no `<scriptfile>` at all. The correlation is exact: `plg_system_cwmscripture`, the one shared extension it does *not* declare, is the only one that survives.

⚠️ The guard also **reported success while failing**. Phase 6 installs `pkg_proclaim`, which bundles `pkg_cwmscripture` and re-parents the stack, so by phase 10 it found nothing to fix and printed *"already owned by pkg_cwmscripture"* — correct ownership — immediately before three extensions were destroyed anyway.

## What replaces it

Reinstall `pkg_cwmscripture` after the clearing loop. This restores whatever was taken and works for **any** consumer whose manifest overlaps ours, not just LivingWord.

- The version comes from `pkg_cwmscripture.xml`, not a glob — several builds accumulate in `build/dist`, and taking the newest file is how a package once shipped a zip that was not the pinned one.
- The assertion sits **immediately after the restore**, not at the next phase's first use. Every phase from 11 on needs a library, so without it the damage surfaces as a missing fixture reported against whichever assertion ran first — which is how this cost a full diagnostic pass to attribute.

## Verification

Reproduced the damage on j6-test with `pkg_livingword` installed — `extension:remove` on the package took the library and both plugins — then ran the added block verbatim:

```
   restoring the scripture stack those removals may have taken:
     reinstalled pkg_cwmscripture 1.2.10
  OK   lib_cwmscripture is still installed — the uninstall was refused
```

All three return, parented to `pkg_cwmscripture`.

⚠️ **A full 17-phase run with LivingWord present is not yet green**, for a reason upstream of this change: phase 9 fails with the seeded translation, verses and cache destroyed during the upgrade. Being tracked separately — it is not caused by, nor fixed by, this PR, which only touches phase 10 onward. The full run passes with no consumer installed.

### DIFF
--- a/build/test-upgrade.sh
+++ b/build/test-upgrade.sh
@@ -135,21 +135,20 @@ fi
 OTHER_CONSUMERS="$(php build/verify-scripture-uninstall.php other-consumer-ids | grep -E '^[0-9]+$' || true)"
 
 if [ -n "$OTHER_CONSUMERS" ]; then
-    # Before removing anything: hand the shared scripture stack back to
-    # pkg_cwmscripture. The library and plg_content_scripturelinks are package
-    # children, and package_id points at whichever package installed them last
-    # -- on a site carrying LivingWord that can be pkg_livingword, so removing
-    # it below took the library and the content plugin with it. Every phase
-    # after this then failed on a missing library, and because these removals
-    # persist, so did the next run (#1820).
-    echo "   protecting the scripture stack from the removals below:"
-
-    if ! php build/verify-scripture-uninstall.php protect-scripture-ownership; then
-        echo "ERROR: could not protect the scripture stack; refusing to remove consumers." >&2
-        echo "       Removing them now could take lib_cwmscripture with them." >&2
-        exit 1
-    fi
-
+    # ⚠️ Removing a consumer package takes the shared scripture stack with it,
+    # and correcting package_id cannot prevent that. PackageAdapter::
+    # removeExtensionFiles() reads the consumer's *installed* manifest from
+    # administrator/manifests/packages and resolves every <file> it declares
+    # through _getExtensionId($type, $element, $client, $group) -- by element.
+    # package_id is never consulted on that path.
+    #
+    # pkg_livingword declares lib_cwmscripture, plg_content_scripturelinks and
+    # plg_task_cwmscripture as its own children, so all three go with it.
+    # plg_system_cwmscripture, which it does not declare, survives -- the
+    # correlation is exact (#1860, and the same damage as #1820).
+    #
+    # So the stack is restored afterwards rather than defended beforehand. That
+    # holds for any consumer whose manifest overlaps ours, not just LivingWord.
     echo "   clearing other scripture consumers before the uninstall phases:"
 
     for CID in $OTHER_CONSUMERS; do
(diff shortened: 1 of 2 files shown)


bcordis:
## Now verified end to end, with a consumer present

Reinstalled `pkg_livingword` **5.7.0** on j6-test and ran the full harness: **`UPGRADE TEST PASSED 10.5.9 -> 10.5.10-dev`**, exit 0, all 17 phases.

Phase 10 doing the work:

```
   clearing other scripture consumers before the uninstall phases:
     removed extension id 4417
   restoring the scripture stack those removals may have taken:
     reinstalled pkg_cwmscripture 1.2.10
  OK   lib_cwmscripture is still installed — the uninstall was refused
```

The consumer is removed, it takes the stack with it exactly as before, the reinstall puts it back, and the assertion confirms it — all inside the phase that used to leave the wreckage behind. As far as I can tell this is the **first green run of this harness with a scripture consumer installed**; phases 11–17 have only ever run on a site with no consumer, where phase 10's destructive branch is skipped entirely.

## Withdrawing the phase-9 caveat in the description

The earlier run failed at phase 9 with the seeded translation, verses and cache destroyed. That was **specific to the `pkg_livingword-5.7.0-beta4` build**, not a Proclaim defect:

- Installing 5.7.0 leaves `libraries/cwmscripture/sql/uninstall.mysql.utf8.sql` at 957 bytes with no non-comment content — disarmed, unchanged before and after, library still 1.1.18. It does not re-arm the uninstall SQL.
- With 5.7.0 installed, phase 9 passes: probe translation still registered, all 5 verses and all 3 cache rows survived.

The beta4 zip was replaced on disk mid-investigation, so the armed-file check on that specific build is no longer possible. Nothing to file against Proclaim; noting it here in case the same symptom resurfaces on another consumer build.

Issue #102, opened 2013-03-06
Podcast xml files not writing and bug with custom titles


tomfuller2:
See here:
http://www.joomlabiblestudy.org/forum/User-Assistance/3545-custom-podcast-title.html#3721

On version 8, I get no message on WriteXML and no files written.

This is a Internal Dev bug. and This should not be included into the change Order Log.

Issue #1879, opened 2026-08-17
Demo study alias is a 264-character binary string, silently truncated to 255 on install


bcordis:
The seeded demo study in `admin/sql/install.mysql.utf8.sql` has an `alias` that
is the *binary encoding* of its intended slug rather than the slug itself.

**Row:** `#__bsms_studies` `id = 1`, studytitle `Four Steps to Defeating the Flesh`

The stored value is 264 characters of `0` and `1`:

```
0110011001101111011101010111001000101101011100110111010001100101...
```

Decoded as 8-bit groups, that is exactly:

```
four-steps-to-defeating-the-flesh
```

33 characters × 8 bits = 264 characters.

### Why nobody has noticed

`#__bsms_studies.alias` is declared `VARCHAR(255)`, and
`install.mysql.utf8.sql` opens with `SET SQL_MODE = ''`. Non-strict mode means
MySQL silently truncates the 264-character value to 255 instead of erroring.
So every install ships a demo study whose URL slug is 255 characters of binary
digits — and the truncation drops the last 9 bits mid-byte, so it does not even
decode back to a whole string:

```
first 255 chars decode to: "four-steps-to-defeating-the-fle" + 7 orphan bits
```

### How it surfaced

PostgreSQL has no non-strict mode. Loading the converted schema fails hard:

```
ERROR:  value too long for type character varying(255)
```

### Fix

Set the alias to the literal slug:

```sql
'four-steps-to-defeating-the-flesh'
```

### Notes

- This is a live bug on MySQL today, not a PostgreSQL-only concern. It's worth
  fixing independently of whether PostgreSQL support ever ships.
- Existing sites already hold the truncated binary value, so a data-repair
  migration is worth considering alongside the seed fix.
- Worth auditing the rest of the seed data for the same class of overflow —
  `SET SQL_MODE = ''` hides all of them equally.

bcordis:
Surfaced while converting the schema for #1857, but this is a live MySQL bug and independent of whether PostgreSQL support ever ships.

bcordis:
Fixed in #1884, with one correction to the diagnosis above.

**This is not a live bug on MySQL, and no data-repair migration is needed.**

The value is a MySQL *hex literal*, not a string of binary digits:

```sql
X'666F75722D73746570732D746F2D646566656174696E672D7468652D666C657368'
```

[...]

Issue #1755, opened 2026-08-12
The API acceptance job does not run when the API's models change


bcordis:
`.github/workflows/e2e.yml` gates the **Package install + API acceptance** job on these paths:

```
api/**
plugins/webservices/**
tests/e2e/**
build/…
playwright.config.js
.github/workflows/e2e.yml
```

`admin/src/Model/**` is not among them. **But the API has no models of its own** — that is the design. Every list route dispatches to an admin `Cwm*Model`, resolved through each controller's `getModel()` map. The list query the API returns *is* the admin model's `getListQuery()`.

So a change that breaks every API list response does not trigger the suite that would catch it.

## Demonstrated, not theorised

PR #1754 (#1753) changed `CwmcommentsModel`, `CwmplaylistsModel`, `CwmserversModel` and `CwmtopicsModel` — four models the API dispatches to, altering the WHERE clause of four list endpoints. Seven checks ran. **Package install + API acceptance was not one of them.** PR #1752, which touched `api/**`, did run it.

The same gap covers the earlier work: #1752's `LIST_SELECT` fix depended on `list.select` behaving as the admin models implement it, and #1743/#1750 changed model queries directly.

## The awkward part

The blunt fix — adding `admin/src/Model/**` — would run a ~3-minute package install and Playwright suite on a large share of PRs, since those models change often. Worth weighing against the cost of an undetected API regression, which is what actually happened three times in 10.3.0–10.3.3 (#1309/#1310/#1330).

Options:

1. **Add `admin/src/Model/**`.** Correct and simple; costs CI minutes on many PRs.
2. **List only the twelve models the API dispatches to.** Cheaper to run, brittle to maintain — and nothing would fail if a thirteenth were added.
3. **Run it on every PR to `development`, drop the path filter.** Simplest to reason about; highest cost.
4. **Leave the filter, add a unit-level guard** asserting the API's twelve models are named in the workflow's path list. Cheap, and it makes option 2 self-maintaining.

2 + 4 together look like the best trade — the granularity of 2 without its main failure mode.

## Related

- #1749 / PR #1752 — response fields that were never in a response
- #1753 / PR #1754 — list filters accepted and discarded
- Both were found by reading code, not by CI.

PR #1892, opened 2026-08-18
refactor(build): migrate verify-frontend to TestSite, fixing the same second-install fatal

bcordis:
Fourth file off hand-rolled `mysqli` (#102), and the **third** carrying the
`JConfig` redeclare — same shape as #1891:

```php
[$host, $user, $pass, $name, $prefix] = (static function (string $file): array {
    require $file;
    $c = new \JConfig();
    ...
})($configFile);
```

inside `foreach ($installs as $install)`. Classes are process-global, so the
second `role=test` install is fatal. Demonstrated by declaring one:

| | result |
|---|---|
| before | **exit 255**, dies after printing the second install's header |
| after | **exit 0**, both installs verified |

That's three of the four migrated files with the same latent fatal, which is a
decent argument that the primitive was worth extracting.

## Also fixed: a failure that told you to do the wrong thing

`mysqli_query()` returns `false` on error, and the old loop read that as an
empty result set:

```php
while ($result && $row = mysqli_fetch_assoc($result)) {
```

A failed query therefore reported *"no seeded menu items — run
build/seed-testsite-menus.php first"* — sending the reader off to reseed a site
whose query had actually failed. PDO with `ERRMODE_EXCEPTION` cannot produce
that outcome. Same class of defect as the rest of #102: a check that passes, or
fails, while testing something other than what it claims.

The menu lookup also binds `SEED_NOTE` rather than interpolating it. It is a
project-owned constant so this was never exploitable — it just matches the
shape the other migrated files now use.

## Verified

Byte-identical against the live `j6-test` site in **both** states:

| state | result |
|---|---|
| unseeded | exits 1 on the empty menu — identical |
| seeded (`build/seed-testsite-menus.php`) | all 4 page checks pass, exit 0 — identical |

The seeded run matters because it exercises both queries. The reference lookup
resolves `Colossians 3:5-11`, so `$expectedBook` is non-null and the
Books-section assertion that depends on it actually runs instead of being
skipped.

`php-cs-fixer` clean locally before pushing: 0 of 617.

Refs #102

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -60,6 +60,24 @@ on:
       - 'build/build-subpackages.sh'
       - 'build/verify-api-install.php'
       - 'build/verify-migrations.php'
+      # ⚠️ test-install.sh and test-upgrade.sh invoke twelve build/ scripts
+      # between them, and only four were listed here. The other eight were
+      # executed by this job on every run while never triggering it, so a
+      # change to one reached CI only when it happened to ride along with a
+      # gated file. verify-frontend.php was migrated off mysqli in #1892 and
+      # this job -- the only thing that runs it -- did not start. Same shape as
+      # #1755, #1754 and #1879 above.
+      #
+      # These are build-harness scripts that change rarely, so unlike the
+      # admin/src entries above this costs almost nothing in extra runs.
+      - 'build/build-baseline.sh'
+      - 'build/seed-testsite-menus.php'
+      - 'build/verify-frontend.php'
+      - 'build/verify-install-log.php'
+      - 'build/verify-schema-check.php'
+      - 'build/verify-scripture-install.php'
+      - 'build/verify-scripture-uninstall.php'
+      - 'build/verify-scripture-upgrade.php'
       - 'playwright.config.js'
       # The harness resets the site through cwm-reset-testsite and resolves its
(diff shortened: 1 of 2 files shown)


bcordis:
## Second commit: the reason this PR's own job didn't run

The `Package install + API acceptance` job did not trigger on the first push
here. It should have — that job is the **only** thing that runs
`verify-frontend.php`, via `build/test-install.sh`.

`test-install.sh` and `test-upgrade.sh` invoke **twelve** `build/` scripts
between them. Four were in the `paths:` allow-list:

```
build/test-install.sh          build/build-package.sh
build/build-subpackages.sh     build/verify-api-install.php
build/verify-migrations.php
```

The other eight were executed on every run while never triggering one:

```
build/build-baseline.sh              build/seed-testsite-menus.php
build/verify-frontend.php            build/verify-install-log.php
build/verify-schema-check.php        build/verify-scripture-install.php
build/verify-scripture-uninstall.php build/verify-scripture-upgrade.php
```

`paths:` fails closed, so the symptom is silence. `verify-scripture-install.php`
was migrated in #1890 and only reached this job because that PR happened to
also touch `verify-api-install.php`.

Same shape as #1755, #1754 and #1879 — each already documented in this file's
comments. Added all eight; every path entry in the workflow now matches a real
file (25 entries, none stale).

